PR #1, opened 2019-07-06
qmail.o rule is missing dependency on env.h via qmail.c.

alanpost:


schmonz:
Agreed.

Even though this PR is safe and extremely small, I'd like to defer merging. This is a great chance to practice (and set precedent for) how we're going to handle proposed changes to existing qmail code.

We currently have "Modifying or replacing core components" as an explicit [non-goal](https://github.com/notqmail/notqmail/wiki#non-goals). This non-goal was set by me unilaterally, without discussion or consensus. I'd like us to decide whether it's a good non-goal and whether we've stated it clearly enough (since I meant it to apply to this case, but it doesn't exactly sound like it does). And since our project just got started today, I'd like to wait until "us" can include a few more of the people I've invited.

alanpost:
If there was going to be a netqmail-1.07 this patch would be in it.  The patch fixes a minor bug, introduced in netqmail-1.04.  It is only triggered by an incremental rebuild with a qmail.c facing change to env.h (e.g., env_get).  It truth netqmail has never modified env.h.  Therefor a series of netqmail builds incrementally for each release would not have triggered this issue.

The limits of the issue defined, this is strictly a correctness improvement: the Makefile DAG is incorrect, this patch corrects it.  I searched for this class of build problem (missing dependencies) exhaustively and it is the only extant example in the netqmail codebase.

> We currently have "Modifying or replacing core components" as an explicit non-goal.

One literally has to invent a reason not to accept this patch.  It corrects a simple, demonstrable, verifiable mistake and has no side-effects.  Don't get confused by your own ideals (goals) when faced with an issue that can be decided on the evidence, empirically.

schmonz:
Let me be clear: I want this patch to go in. (Also, when we have some automated tests and/or autobuilders, I'd love to weaponize your search for missing header dependencies and add it as a red/green test.)

I also want this patch, which is clearly conservative and correct, to prompt us to clarify our goals and non-goals, so that they accurately reflect which sorts of changes to "core qmail" we do and don't want. I think we all agree this should go in. I'm thinking ahead to the ones we might not agree on, and whether our stated intentions will provide us helpful guidance at such times.

Now that I've said that out loud, I no longer feel such a strong need to have a clarifying discussion about goals and non-goals before we've even encountered one where we can't manage to agree. We all agree on this one. I'm sure we'll have chances to discuss later. :-)

PR #3, opened 2019-07-07
destdir: build as non-root, without hardcoded IDs.

schmonz:
This is my destdir patch from <https://schmonz.com/qmail/destdir/>. With
it, binary packages built on one system can run on any other system of
matching OS and architecture without also requiring matching numeric
UIDs and GIDs.

This patch supports our explicit goal "Being easily packaged by OS
integrators". I added it to pkgsrc 2 years ago, and they've had working
binary qmail packages ever since. My production mail server runs one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,14 @@ SHELL=/bin/sh
 
 default: it
 
+auto_destdir.c: \
+auto-str conf-destdir
+	./auto-str auto_qmail `head -1 conf-destdir` > auto_destdir.c
+
+auto_destdir.o: \
+compile auto_destdir.c
+	./compile auto_destdir.c
+
 addresses.0: \
 addresses.5
 	nroff -man addresses.5 > addresses.0
@@ -110,7 +118,7 @@ auto_split.o: \
 compile auto_split.c
 	./compile auto_split.c
 
-auto_uids.c: \
+auto_uids_orig.c: \
 auto-uid auto-gid conf-users conf-groups
 	( ./auto-uid auto_uida `head -1 conf-users` \
 	&&./auto-uid auto_uidd `head -2 conf-users | tail -1` \
@@ -122,10 +130,10 @@ auto-uid auto-gid conf-users conf-groups
 	&&./auto-uid auto_uids `head -8 conf-users | tail -1` \
 	&&./auto-gid auto_gidq `head -1 conf-groups` \
 	&&./auto-gid auto_gidn `head -2 conf-groups | tail -1` \
-	) > auto_uids.c.tmp && mv auto_uids.c.tmp auto_uids.c
+	) > auto_uids_orig.c.tmp && mv auto_uids_orig.c.tmp auto_uids_orig.c
 
 auto_uids.o: \
-compile auto_uids.c
+compile auto_uids.c auto_usergroupnames.h
 	./compile auto_uids.c
 
 auto_usera.c: \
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -136,6 +144,20 @@ auto_usera.o: \
 compile auto_usera.c
 	./compile auto_usera.c
 
+auto_usergroupnames.h: \
+auto-str conf-users conf-groups
+	( ./auto-str auto_usera `head -1 conf-users` \
```
DerDakon: This should be "tail -n 1" for extended Posix compatibility.

Makefile
```diff
@@ -4,6 +4,14 @@ SHELL=/bin/sh
 
 default: it
 
+auto_destdir.c: \
+auto-str conf-destdir
```
DerDakon: I fear this doesn't really make anything better for packagers. They like real staging directories, passed by DESTDIR environment variable.

Makefile
schmonz: Hmmm, I don't see `tail -n` anywhere else in `Makefile`. Pretty sure I copy-pasted the form followed elsewhere in the file. If this is an important family of changes to make (maybe because it fixes broken builds on some platform?), please propose it as a separate PR.

Makefile
schmonz: I agree that third-party software supporting `${DESTDIR}` is very friendly to packagers. As a packager, I can also attest that this change (and similar ones I made in ucspi-tcp, djbdns, et al.) solved the staging problem just as well as `${DESTDIR}` would have, and I think less invasively.

Makefile
DerDakon: If things are that way in other places I'm ok with it for now.

Makefile
DerDakon: It just means the Makefile is worse than the average. Probably just it's age.

auto_uids.h
```diff
@@ -1,16 +1,29 @@
...
 #define AUTO_UIDS_H
 
-extern int auto_uida;
-extern int auto_uidd;
-extern int auto_uidl;
-extern int auto_uido;
-extern int auto_uidp;
-extern int auto_uidq;
-extern int auto_uidr;
-extern int auto_uids;
```
alanpost: Temperature check for the team.  These variables can be set at runtime via running a procedure using the init section of an ELF executable.  Example:

    #include <sys/types.h>
    #include <pwd.h>
    #include <stdlib.h>
    #include <stdio.h>

    static uid_t root_uid = -1;
    static void lookup(void) __attribute__((constructor));

    static void
    lookup(void)
    {
        struct passwd *pw;
        pw = getpwnam("root");
        if(!pw) { exit(1); }
        root_uid = pw->pw_uid;
    }

    int
    main(int argc, char *argv[])
    {
        printf("root uid: %ld\n", (long)root_uid);
        exit(0);
    }

which produces the output:

    root uid: 0

Here the procedure lookup, due to the `__attribute__((constructor))` directive, is added to the .init section of the executable and run before main.  This is the mechanism that C++ uses for static initialization.

If we used this feature we'd be able to retain the `auto_` variables here, deferring when they are populated to runtime.  Further we could use different values in different executable programs, so this same feature could be used for the fake_chown work this patch does.

Overall I think using .init to populate these variables would lead to a simpler patch, at the expense of introducing a new compiler construct to the code.  Said construct is widely available, owing to C++ requiring it, so I believe the cost to be modest and worth using here, but I'm interested in what everyone thinks about this alternative.

schmonz: This is cool! TIL. My main worry is portability. What happens on non-ELF systems like OS X? How recent a compiler would we be requiring?

alanpost: Here is the same program on OS X, courtesy of hmadison.  This construct works on OS X, producing the same output I report above on Linux.

https://gist.github.com/hmadison/4ecab14ddbcf3bd651ab848e6d578d53

The test included in it's entirety here:

    hmadison@hunters-mbp ~/s/test> cat alynpost.c 
    #include <sys/types.h>
    #include <pwd.h>
    #include <stdlib.h>
    #include <stdio.h>

    static uid_t root_uid = -1;
    static void lookup(void) __attribute__((constructor));

    static void
    lookup(void)
    {
        struct passwd *pw;
        pw = getpwnam("root");
        if(!pw) { exit(1); }
        root_uid = pw->pw_uid;
    }

    int
    main(int argc, char *argv[])
    {
        printf("root uid: %ld\n", (long)root_uid);
        exit(0);
    }
    hmadison@hunters-mbp ~/s/test> gcc alynpost.c 
    hmadison@hunters-mbp ~/s/test> ./a.out 
    root uid: 0
    hmadison@hunters-mbp ~/s/test> 
    hmadison@hunters-mbp ~/s/test> uname -a
    Darwin hunters-mbp.lan 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64

DerDakon: This sounds useful, but this should be discussed in it's own issue or PR, or maybe on some mailing list.

schmonz: BTW, I can't remember why I bundled "looking up IDs at runtime" and "staged installs" into a single patch -- maybe because deferring IDs at runtime let me do staged installs as a non-root user? -- but if we can unbundle the two features and evaluate implementations of each on its own merits, we probably should.

alanpost: > This sounds useful, but this should be discussed in it's own issue or PR, or maybe on some mailing list.

Fully agreed.  I've submitted PR#11 that covers the runtime lookup of uid/gid in the manner I describe here.

schmonz:
This PR has two ideas bundled together. We are addressing them separately:

- #15 for building as non-root
- #4 (already merged) for installing into a staging directory

PR #4, opened 2019-07-08
honor DESTDIR on install

DerDakon:
In contrast to #3 this does not need any new config files, it will just work as expected to packagers. The uid/gid masking is left out here to have the features discussed one by one.

### DIFF
--- a/install.c
+++ b/install.c
@@ -1,9 +1,14 @@
 #include "substdio.h"
 #include "strerr.h"
+#include "env.h"
 #include "error.h"
+#include "fifo.h"
 #include "open.h"
 #include "readwrite.h"
 #include "exit.h"
+#include "alloc.h"
+#include "str.h"
+#include "stralloc.h"
 
 extern void hier();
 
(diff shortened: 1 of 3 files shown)


install.c
```diff
@@ -4,26 +4,73 @@
 #include "open.h"
 #include "readwrite.h"
 #include "exit.h"
+#include "alloc.h"
+#include "stralloc.h"
 
 extern void hier();
+extern char *getenv(const char *name);
```
alanpost: In the qmail codebase, getenv is env_get, requiring the header env.h and linking against env.a.

DerDakon: Good catch. It was also missing fifo.h (DJBs fault, he deleted it in 1.03) and str.h (for str_len(), my fault).

install.c
```diff
@@ -1,29 +1,78 @@
...
 #define FATAL "install: fatal: "
 
 int fdsourcedir = -1;
 
+static void die_nomem()
+{
+  strerr_die2sys(111,FATAL,"out of memory");
+}
+
+static int ddhome(stralloc *dd, const char *home)
```
alanpost: This routine is called for every installed unit, but the result is the ~same every time (presuming all of the supporting routines always pass `auto_qmail` as `home`).  Should it be memoized?  That could be accomplished by changing where it is called (in main, which would also require a mechanism like fdsourcedir because of the change in calling convention), by using a constructor attribute for the one-time initialization, or by having a static flag for the same.  Thoughts?

install.c
```diff
@@ -1,29 +1,78 @@
...
+{
+  const char *denv = env_get("DESTDIR");
+  if (!denv)
+    return 0;
+
+  if (!stralloc_copys(dd,denv)) die_nomem();
+  if (!stralloc_catb(dd,home,str_len(home))) die_nomem();
```
alanpost: It might be worth adding the following above this line:

    if (home[0] != '/') {
      if (!stralloc_copys(dd,"/")) die_nomem();
    }

In the event that the first line in conf-qmail isn't slash (/).

DerDakon: This has been broken forever, I don't think this could ever work.

install.c
DerDakon: This is true, but it would require some refactoring in hier.c, which calls all these functions with the same prefix every time. This is not needed, it would be enough to pass it once in h(), but I would leave it for the moment to keep the diff small.

install.c
```diff
@@ -1,29 +1,78 @@
...
+  int r = mkdir(home,0700);
+  if (!r)
+    return 0;
+
+  if (errno != error_noent)
+    return r;
+
+  /* try parent first */
+  sl = str_rchr(home, '/');
+  if (!stralloc_copyb(&parent,home,sl)) die_nomem();
```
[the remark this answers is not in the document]
DerDakon: good catch

install.c
alanpost: I'm not sure what you're referring to, as regards what is broken.  As best as I can tell DESTDIR with or without a trailing slash (/) works here, because home (default: /var/qmail) has a leading slash.  Reading through the mkdir_p implementation, I found it to work if DESTDIR had a trailing slash, which would lead to a double slash (//) in `dd`.  That can leading to one or more trailing slashes in the argument passed to mkdir(2), that syscall tolerates that condition.

If you were referring to something else please excuse the noise.  I did find this behavior subtle but couldn't find anything wrong except in the case when conf-qmail is a relative path *and* DESTDIR is set, which is the genesis of my comment about adding a slash in that case.  In practice I doubt this comes up, much as it's worth fixing.

DerDakon: I think that a relative path in conf-qmail would never work, as all programs try to do ```chdir(auto_qmail)``` at the beginning, which would fail for a relative homedir. DESTDIR with a trailing / would not do anything bad, worst is that it would try to create the same directory twice. There was a race condition in that code, which I will remove in a second.

PR #5, opened 2019-07-08
Destdir alt

mbhangui:
Hi @schmonz as suggested by you, I have created a pull request and submitted the method I use for my RPM and debian builds. This allows one to use make DESTDIR=/tmp/abcd install to install /var/qmail hierarchi inside /tmp/abcd

### DIFF
--- a/install.c
+++ b/install.c
@@ -1,15 +1,48 @@
+#include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
 #include "substdio.h"
 #include "strerr.h"
+#include "fifo.h"
 #include "error.h"
 #include "open.h"
-#include "readwrite.h"
 #include "exit.h"
+#include "stralloc.h"
+#include "sgetopt.h"
 
 extern void hier();
 
 #define FATAL "install: fatal: "
 
 int fdsourcedir = -1;
+uid_t my_uid;
+gid_t my_gid;
+char *destdir = 0;
+stralloc tmpdir = { 0 };
+stralloc dirbuf = { 0 };
+
+int
+r_mkdir(home, mode)
+	char           *home;
+	int             mode;
+{
+	char           *ptr;
+	int             i;
+
+	if (!stralloc_copys(&dirbuf, home))
+		strerr_die2sys(111, FATAL, "out of memory: ");
+	if (!stralloc_0(&dirbuf))
+		strerr_die2sys(111, FATAL, "out of memory: ");
+	for (ptr = dirbuf.s + 1;*ptr;ptr++) {
+		if (*ptr == '/') {
+			*ptr = 0;
+			if (access(dirbuf.s, F_OK) && (i = mkdir(dirbuf.s, mode)) == -1)
+				return (i);
+			*ptr = '/';
+		}
+	}
+	return (mkdir(dirbuf.s, mode));
+}
 
 void h(home,uid,gid,mode)
 char *home;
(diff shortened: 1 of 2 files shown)


DerDakon:
This is basically what I have cooked up today in #4, too. My patch is a bit smaller as it avoids the command line parsing and just uses DESTDIR from the environment, and the code in every function is reduced as I have moved things into a function.

schmonz:
Thanks @mbhangui for being willing to try the GitHub way of doing things. I still remember when I was first learning Git and GitHub, coming from CVS. It's a lot to adjust to (and I'm still adjusting!).

One thing I see is, it looks like the branch you made for this PR is directly in the central notqmail repo. That works, but it's much preferable to have your own notqmail "fork" where you make branches for PRs. For your next PR, let's share a screen and submit it together.

DerDakon:
I have done the same ;)

schmonz:
Maybe it's my limited experience limiting me, then. I've only used fork+branch+PR. People who aren't yet part of our project will have to fork first, of course. But is it better/easier/just-as-safe for us if we make PR branches directly in the central notqmail repo?

DerDakon:
Both works. I have set up some branch protection rules so we should be reasonably safe from accidential damage. Everything else is policy and should be discussed elsewhere, not in this PR.

install.c
```diff
@@ -1,28 +1,71 @@
...
+	int             i;
+
+	if (!stralloc_copys(&dirbuf, home))
+		strerr_die2sys(111, FATAL, "out of memory: ");
+	if (!stralloc_0(&dirbuf))
+		strerr_die2sys(111, FATAL, "out of memory: ");
+	for (ptr = dirbuf.s + 1;*ptr;ptr++) {
+		if (*ptr == '/') {
+			*ptr = 0;
+			if (access(dirbuf.s, F_OK) && (i = mkdir(dirbuf.s, mode)) == -1)
```
DerDakon: there is a race condition here: if someone creates the directory between the access() and the mkdir() the function will fail. You need to check errno == error_exist after the mkdir().

install.c
```diff
@@ -1,28 +1,71 @@
...
+				return (i);
+			*ptr = '/';
+		}
+	}
+	return (mkdir(dirbuf.s, mode));
```
DerDakon: needless outer ()

install.c
```diff
@@ -127,7 +201,17 @@ int mode;
...
   int fdout;
 
-  if (chdir(home) == -1)
+  if (destdir) {
+    if (!stralloc_copys(&tmpdir, destdir))
+      strerr_die2sys(111, FATAL, "out of memory: ");
+    if (!stralloc_cats(&tmpdir, home))
+      strerr_die2sys(111, FATAL, "out of memory: ");
+  } else
+  if (!stralloc_copys(&tmpdir, home))
```
DerDakon: The indentation is misleading here.

install.c
```diff
@@ -146,17 +230,35 @@ int mode;
...
 
-  if (chown(file,uid,gid) == -1)
+  if (!my_uid && chown(file,uid,gid) == -1)
     strerr_die6sys(111,FATAL,"unable to chown ",home,"/",file,": ");
   if (chmod(file,mode) == -1)
     strerr_die6sys(111,FATAL,"unable to chmod ",home,"/",file,": ");
 }
 
-void main()
+char           *usage = "usage: setup -d destdir";
```
DerDakon: no need for that variable, just put the string to where it is used.

alanpost:
@mbhangui,

We went with pr#4, @DerDakon's DESTDIR implementation.  That it worked via directly fetching the value from the environment, and didn't require a command-line option or the shell (via make in your patch), making it the more compelling implementation.

I very much enjoyed your story on the dev mailing list, and look forward to further pull requests.

mbhangui:
Also @DerDakon code was shorter. But the reason why I had it as command line was because it is implemented that way in indimail-mta for full FHS compliance. I have conf-sysconfdir, conf-logdir for specifying the directory for confg files & the directory for the supervise log files. e.g. something like this

install: \
it man leapsecs.dat conf-sysconfdir perm_list.q conf-logdir
    if test ! -d $(DESTDIR)"`head -1 conf-sysconfdir`" ; then \
        mkdir -p $(DESTDIR)"`head -1 conf-sysconfdir`" ; \
    fi ;
    if test ! -d $(DESTDIR)"`head -1 conf-logdir`" ; then \
        mkdir -p $(DESTDIR)"`head -1 conf-logdir`" ; \
    fi ;
    ./setup -d "$(DESTDIR)" -L $(DESTDIR)"`head -1 conf-logdir`" \
        $(QMAILDIR) && \
        cp perm_list.q $(DESTDIR)"`head -1 conf-sysconfdir`"

NOTE: I have renamed install.c as setup.c to avoid it clashing with make install command

PR #11, opened 2019-07-09
detect qmail system uid/gid at runtime.

alanpost:
rename and extend auto_usera.h to auto_users.h and
include all system accounts, not just the alias
user.

Instead of looking up the uid/gid of each account
in conf-users at build time, record their names
in the same manner as the alias user.

remove auto-gid.c and auto-uid.c, replacing them
with runtime equivalent gid.c and uid.c  Using the
.init section (via the constructor attribute),
set the uid and gid values in auto_uids.h before
main, using the newly defined account names.

Separately compile the runtime lookup for each uid
and gid and arrange to link only those used by
each program.

Link substdio.a in to qmail-start, as runtime error
messages are produced when an account cannot be
found.

This allows qmail to be built on a host that does
not have user and group system accounts.


alanpost:
This PR is superseded by PR#15, which includes the patch here along with others that build on it.

PR #12, opened 2019-07-09
Fix build on utmpx-only systems, such as FreeBSD.

schmonz:
FreeBSD removed <utmp.h> almost a decade ago, leaving only <utmpx.h>:
https://github.com/freebsd/freebsd/commit/cb38fdae7c004eec078edf577e1819ce9760aecb

This is my qbiff-utmpx patch from <https://schmonz.com/qmail/qbiffutmpx/>.
With it, on a system...

- with <utmpx.h> but no <utmp.h>: build proceeds past qbiff.c

- with <utmpx.h> and <utmp.h>: build chooses <utmpx.h>

- with no <utmpx.h>: no change in behavior

Supports goals: being easily packaged by OS integrators
Risks: may run qbiff incorrectly on a system with broken utmpx implementation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -685,6 +685,13 @@ chkshsgr warn-shsgr tryshsgr.c compile load
 	hasshsgr.h
 	rm -f tryshsgr.o tryshsgr
 
+hasutmpx.h: \
+tryutmpx.c compile load
+	( ( ./compile tryutmpx.c && ./load tryutmpx ) >/dev/null \
+	2>&1 \
+	&& echo \#define HASUTMPX 1 || exit 0 ) > hasutmpx.h
+	rm -f tryutmpx.o tryutmpx
+
 haswaitp.h: \
 trywaitp.c compile load
 	( ( ./compile trywaitp.c && ./load trywaitp ) >/dev/null \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -685,6 +685,13 @@ chkshsgr warn-shsgr tryshsgr.c compile load
 	hasshsgr.h
 	rm -f tryshsgr.o tryshsgr
 
+hasutmpx.h: \
+tryutmpx.c compile load
+	( ( ./compile tryutmpx.c && ./load tryutmpx ) >/dev/null \
+	2>&1 \
+	&& echo \#define HASUTMPX 1 || exit 0 ) > hasutmpx.h
```
DerDakon: I would additionally add the ```#include <utmpx.h>``` here, so one does not need another ```#ifdef``` after including this file. At the end this would end up with 2 includes like it is done e.g. with fork.h.

tryutmpx.c
```diff
@@ -0,0 +1,6 @@
+#include <utmpx.h>
+
+void main()
+{
+  ;
```
DerDakon: this should probably also include the usage of struct utmpx, ideally probing those members that are actually used in qbiff (i.e. ut_line, ut_type, and ut_user).

DerDakon:
Please rebase that on current master so a CI run is triggered.

schmonz:
I let GitHub talk me into removing my "unused" fork after another PR was merged, but this PR had been based on a branch there and I can't figure out how to repoint the PR at the (identically-named) branch directly in notqmail, even though browsing to notqmail's "utmpx" branch does show it to be correlated with this PR. Closing this one so I can resubmit with my updated code.

schmonz:
Superseded by #24.

PR #13, opened 2019-07-09
add .gitignore

DerDakon:
This is basically the result of

```shell
head -n -3 TARGETS > .gitignore
```


alanpost:
I pushed but didn't issue a PR for a similar change:

https://github.com/alanpost/notqmail/commit/62265b3ddd6342b5fad7786e8ef6687b86eef480

On my branch, I symlinked TARGETS to .gitignore.

If not a symlink, both must be kept in sync.  Also though permits adding files to one but not the other.

alanpost:
PR #25 is the alternative to this.  It creates a symlink of TARGETS to .gitignore.

alanpost:
PR #13 is superseded by PR #25.

PR #15, opened 2019-07-09
split install program in to build host and install host steps removing build time requirement for qmail system accounts

alanpost:
detect the uid/gid of qmail system accounts at runtime, and split the install in to a package (instpackage) and post-install (instchown) steps.  This allows building and packaging as a non-root user by moving the root-requiring steps to programs run on the install host.


auto_uida.c
```diff
@@ -0,0 +1,11 @@
+#include "uidgid.h"
```
DerDakon: These files should be autogenerated from a template I guess, they are just too similar.

alanpost: These are 8+2 files that are self-similar.  I did remove the code in common to uid.c + gid.c, respectively, but it is 3 logical lines repeated 8+2 times.

Do you have a concrete suggestion for the kind of concision you'd like to see here?

I could create a new header file, modeled after gen_allocdefs.h and say something like `INITUID(auto_uid, auto_usersa)`.  Or I could use a mechanism similar to how binm*, config, and datemail are generated.  No C files are generated in this way, currently, as C code generation is done in the manner of auto-str.c or auto-int.c.  But I believe the later here is more what you're thinking.

DerDakon: I would generate the .c files, they are too dump to keep around ;)

CHANGES
```diff
@@ -1,3 +1,7 @@
+20190710 code: add instchown to set ownership and mode on installed files.
+20190709 code: add instpackage to install without root access.
+20190710 code: removed idedit.
```
schmonz: I'd either (1) keep it for now, or (2) remove the whole horse it rode in on.

Argument for keep-for-now: conservatism. "Delete unused code" is generally [slotted for later](https://github.com/notqmail/notqmail/wiki/Release-Plans#maybe-later). If the `idedit` code were broken by your changes, the chance to avoid paying maintenance on unwanted code would make a compelling exception. But it's merely as useless as before, not broken.

Argument for entire-horse: When we claim "Being easily packaged by OS integrators" is a goal, none of us means [DJB's var-qmail packages](https://cr.yp.to/qmail/var-qmail.html). As a successor to qmail, we will need to send clear signals to OS integrators about what we do mean. One very clear signal: our first release is entirely dedicated to providing defaults that allow building as non-root, installing to a staging area, and producing FHS-compatible binary packages. A second clear signal would be to delete everything else related to DJB's var-qmail package scheme, which I believe amounts to these:

- `BIN.Makefile`
- `BIN.README`
- `install-big`
- `config-fast`

Makefile
```diff
@@ -136,6 +117,118 @@ auto_usera.o: \
 compile auto_usera.c
 	./compile auto_usera.c
 
+auto_uida.o: \
+compile auto_uida.c uidgid.h auto_uids.h auto_users.h
+	./compile auto_uida.c
+
+auto_userd.c: \
+auto-str conf-users
+	./auto-str auto_userd `head -2 conf-users` > auto_userd.c
```
schmonz: All the generated `auto_user*.c` are identical (!), getting the first line from `conf-users` (likewise for `auto_group*.c`). In the original sequences, files after the first one had a [`| tail -1`](https://github.com/notqmail/notqmail/pull/15/files#diff-b67911656ef5d18c4ae36cb6741b7965L116), which got lost here. Please re-add.

Makefile
```diff
@@ -728,29 +821,27 @@ hostname.o: \
...
+auto_usera.o auto_uida.o auto_usero.o auto_uido.o auto_userp.o auto_uidp.o \
+auto_userq.o auto_uidq.o auto_userr.o auto_uidr.o auto_users.o auto_uids.o \
+auto_groupq.o auto_gidq.o strerr.a substdio.a open.a error.a str.a fs.a
 	./load install-big fifo.o install.o auto_qmail.o \
-	auto_split.o auto_uids.o strerr.a substdio.a open.a error.a \
-	str.a fs.a 
+	auto_split.o uid.o gid.o auto_usera.o auto_uida.o auto_usero.o \
+	auto_uido.o auto_userp.o auto_uidp.o auto_userq.o auto_uidq.o \
+	auto_userr.o auto_uidr.o auto_users.o auto_uids.o auto_groupq.o \
+	auto_gidq.o strerr.a substdio.a open.a error.a str.a fs.a
```
schmonz: This helps me lean toward whole-horse :-)

Makefile
```diff
@@ -992,6 +1114,10 @@ open_write.o: \
 compile open_write.c open.h
 	./compile open_write.c
 
+package: \
+it man
+	./instpackage
+
```
schmonz: This is great, and gets me thinking about install documentation. The netqmail folks did us the favor of deleting everything in `INSTALL`, leaving nothing but a link to LWQ. I'd say the filename is ours to use.

auto_gidn.c
```diff
@@ -0,0 +1,11 @@
+#include "uidgid.h"
+#include "auto_uids.h"
+#include "auto_users.h"
+
+int auto_gidn;
+
+__attribute__((constructor)) static void
```
schmonz: Reading this, I like how it self-describes, how it works, and how it helps simplify the other changes in this PR.

I worry about compatibility. Besides gcc and clang, which compilers are known to support this? Besides ELF and Mach-O, which executable formats are known to support this? Is this defined in a particular C language standard that we're now requiring?

auto_uida.c
schmonz: +1 to it being a lot of duplication. Usually that worries me more; this code in particular, we probably won't be touching again very often. But if there's a cheap and easy to reduce the duplication, +1 to that too.

gid.c
```diff
@@ -0,0 +1,22 @@
...
+  char *group;
+{
+  struct group *gr;
+  gr = getgrnam(group);
+  if (!gr) {
+    substdio_puts(subfderr,"fatal: unable to find group ");
+    substdio_puts(subfderr,group);
+    substdio_puts(subfderr,"\n");
+    substdio_flush(subfderr);
+    _exit(111);
```
schmonz: When one of the qmail programs starts, fails a lookup, and exits 111, what does that look like? Do we see log entries? As admins, how will we know what happened and what to do about it?

At a glance, looks like the programs to be concerned with are:

- `qmail-lspawn`
- `qmail-queue`
- `qmail-rspawn`
- `qmail-start`

CHANGES
alanpost: I'll pull the commit removing idedit out of this series and bundle it with a new PR removing all of the var-qmail package infrastructure.

config-fast is unusual in your list here.  Gerrit Pape installs that script and runs it in postinst in the netqmail Debian package.  Removing it will break that package.  The ultimate result of our packaging effort will let us broadly remove code from the debian package, as those features will essentially move in-tree.  I'd like to see our packaging work release-ready before making a netqmail package breaking change--so we can make good on that promise of code removal.

As a cross check to the above, the file INSTALL.ctl mentions both config and config-fast.  Removing config-fast.sh naturally brings up the request of also removing config.sh.  config.sh is only available in-tree, uses several supplementary programs (DNS queries), and represents work that needs to be done at install time.  It is certainly true that we support building and installing on the same host, but it is also possible that this use-case will be fully accommodated by the package build pipeline.  This confirms that removing config-fast.sh requires more caution (is a larger change) than idedit or install-big.

I'll submit a patch to remove install-big and idedit (the core of the var-qmail package support) but decline to remove config-fast in favor of deferring that decision to a later PR that better accounts for (at least) Debian packaging and changes to INSTALL.ctl and possibly config.sh.

Whole horse (2), but for config-fast (1).  With one extra bonus removal, stay tuned.  ;-)

DerDakon:
qmail-queue does a lookup of 3 uids (IIRC qmaild, alias and a third one). These are done in order, with alias first This should be changed, as alias is probably the least likely one in usual setups. Then these lookups can be actually dynamic, i.e. been done when the code is hit. It is very likely that at most 2 of them are looked up at all then.

Another way would do to do a lookup for the uid of the calling process, and not lookup the qmail accounts and compare with them. Only if the calling uid can't be mapped to e.g. "qmaild" username, then the backward lookup like it is currently implemented can be done. This may happen e.g. if the uid used for the qmail system users are not unique.

mbhangui:
> qmail-queue does a lookup of 3 uids (IIRC qmaild, alias and a third one). These are done in order, with alias first This should be changed, as alias is probably the least likely one in usual setups. Then these lookups can be actually dynamic, i.e. been done when the code is hit. It is very likely that at most 2 of them are looked up at all then.

The three uids are qmaild, alias and qmails (qmail-send's uid for bounces). Before calling qmail_open(), qmail-smtpd and qmail-send can set one of two bits in a variable to indicate that the caller is qmail-smtpd or qmail-send and call qmail-queue with a command line argument. qmail-queue can avoid password lookups in these two cases. The downside would be a rouge application deliberately and falsely setting this bits to forge the Received headers.

DerDakon:
I also do not think that this lookup is _that_ big deal in runtime costs, so we should just do it properly.

Makefile
alanpost: In that event there is PR #22 removing the shar target and precompiled var-qmail package support.  If that went in before this patch it would remove these lines--but as @DerDakon notes it conceptually fits best right after the first patch in this series, where the smaller idedit removal is now.

I can pull these commits apart again, rehydrating PR #11 to submit the first of these commits, or I can pull in PR #22 to this (already large, I'm afraid) patch series, and I think that gets everyone's needs met.

Makefile
alanpost: Dang, that was sloppy.  Fixed.

auto_uida.c
alanpost: Please have a look at commit a937324.  This is a patch over the top of the patch that makes uid/gid lookup dynamic.  It creates an auto-id program that generates these files rather than generating them by hand and checking them in.

If this is what you both had in mind, I'll squash these two commits together.  Please let me know.

DerDakon: I would have used a auto_id.h.in file with some placeholders and just used sed on them, IIRC other places in the makefile already do this. That avoids building another binary and combining C sources from multiple single strings.

alanpost: Other places in the Makefile do this on shell scripts and manpages, but not for C files, which are generated exclusively by other C programs.  There is a plausible explanation for this, number formatting in auto-int.c and character encoding in auto-str.c.  Yet it is still true that nowhere are we generating C files from shell scripts and sed.

schmonz: On inspection, this feels like one abstraction too far. Pretty sure @DerDakon's suggestion would work for me, and can't think of a reason it'd be a bad idea.

alanpost: I agree with @schmonz, which is why I presented the patch as I originally did.  the code can be fully computed (decided) and the result presented, for the cost of borderline but factually present repetition.  Removing that repetition is too clever by half--it makes things marginally smaller but introduces outsized cognitive overhead.

DerDakon: I had a look in where these things are actually used. Look on everything in showctl and qmail-start: this needs no constructors, it can just have functions doing these lookups. Basically everything but qmail-lspawn does never reuse the uid, so computing it once is just extra code for nothing. So at the end we will not need to split these into _any_ extra code, just throw it all away and have one function that does a lookup for a username string or so.

alanpost: since the constructor attribute is, de-minimus, executing code right before main, it has always been true that this lookup code can be inserted in each main that needs it, eliminating the constructor attribute at the cost of repeating the initialization in all uid/gid using programs.

Let me submit a patch that explores those comments and see if we like the non-attribute using, directly inline at main option--with of course the deferred lookups you've identified for qmail-queue.

gid.c
alanpost: If you run `qmail-start` with a missing user, the program produces the following output:

    qmail-start
    fatal: unable to find user qmaill

The other programs or missing users produce the same result.  Note in the specific case of `qmail-start`, this is an improvement over the prior behavior, where if you built on a system with the qmail system accounts but then ran it on a system where they were not present, it would silently fail.

DerDakon:
I would change all these things to work like this:

```c
uid_t auto_uidX()
{
  static uid_t uid = init_uidX();
  return uid;
}
```

So you do not just use a variable, but call a function to get it. This function will cache the result, so it will do at most one lookup like it is in this patch. It has the benefit of not needing any constructor magic, and that will not needlessly initialize these things until actually required (e.g. for the qmail-queue case).

alanpost:
Team,

have a look at commit f7becbe where I convert all __attribute__((constructor)) calls to initialization of uid/gid values in or about the main function of the five programs plus the installer that use uid or gid values.  If you like this approach I'll squash that commit with the prior one, and we can say goodbye
to both the constructor attribute and it's attendant repetition.

This change does not have the optimization around lookup in qmail-queue--I'm quite pleased to add that but wanted first to get sign-off on this approach as such without unduly complicated the patch.  Please have a look and let me know.

mbhangui:
> Team,
> 
> have a look at commit [02d2455](https://github.com/notqmail/notqmail/commit/02d2455ca76f6c48321736e1963ed30fe943312c) where I convert all `__attribute__((constructor))` calls to initialization of uid/gid values in or about the main function of the five programs plus the installer that use uid or gid values. If you like this approach I'll squash that commit with the prior one, and we can say goodbye
> to both the constructor attribute and it's attendant repetition.

I have checked out this branch on my laptop and taken a look. This is much cleaner without the repetitve files (though I think that the __attribute__((constructor)) calls are cooler).

alanpost:
I have added a commit to this series that reorders the steps in the CI script to verify we can build and do a package install without the qmail system accounts.  It then test that instchown results in an install that passes instcheck, before removing that install and running a regular install.

That tests this feature end-to-end showing build and packaging performed prior to qmail system account creation.

alanpost:
@DerDakon, I have introduced a commit 6725d28 to fix a bug I personally introduced in commit 1dd7b07 *in your name*.  I stated I'd modified the commit, as such, at the bottom of [my test plan for that PR](https://github.com/notqmail/notqmail/pull/4#pullrequestreview-261564047).  I was not aware that I had introduced a bug to your otherwise functional commit.

First, I am sorry.  Second, I am ashamed that I have tarnished your good work.  If you name a retail establishment or other business that you purchase from, I will send you a gift card to atone.

DerDakon:
I don't mind, I guess some of these commits should be squashed or redone anyway.

auto_uida.c
alanpost: qmail-queue does all of it's lookups in `receivedfmt`, which is called up to two times from `received_setup`, so I cannot consistently (i.e., in the same manner as the rest of the programs) set the uid just before use, so I have left the lookups in main.

The constructor attribute feature is now excised from the code.

auto_gidn.c
alanpost: We defined this line of inquiry out of existence by removing the constructor attribute from the code.

Makefile
```diff
@@ -138,6 +111,78 @@ auto_usera.o: \
...
+auto-str conf-groups
+	./auto-str auto_groupn `head -1 conf-groups | tail -1` > auto_groupn.c
+
+auto_groupn.o: \
+compile auto_groupn.c
+	./compile auto_groupn.c
+
+auto_groupq.c: \
+auto-str conf-groups
+	./auto-str auto_groupq `head -2 conf-groups` > auto_groupq.c
```
schmonz: Missing a `| tail -1` again. I’ve always thought it was error-prone that we don’t do `| tail -1` in the `head -1` case. Maybe that would have helped here?

schmonz: Also, I think `groupq` is supposed to be the first line of `conf-groups` and `groupn` the second?

alanpost: I did have that wrong.  It's now fixed.  By hand since this part has been such trouble for me:

Here is the contents of `conf-groups`:

    qmail
    nofiles

    These are the qmail groups. The second group should not have access to
    any files, but it must be usable for processes; this requirement
    excludes the ``nogroup'' and ``nobody'' groups on many systems.

auto_groupq is the first entry, with auto_groupn the second entry.  The tail -1 call needs to be on the second entry in conf-groups, but alphabetically the first group (`n`, not `q`).

    $ head -1 conf-groups
    qmail

    $ head -2 conf-groups | tail -1
    nofiles

Such that the following C files are produced:

    ./auto-str auto_groupn `head -2 conf-groups | tail -1`
    char auto_groupn[] = "\
    \156\157\146\151\154\145\163\
    ";

    ./auto-str auto_groupq `head -1 conf-groups` > auto_groupq.c
    char auto_groupq[] = "\
    \161\155\141\151\154\
    ";

Where `\156` is is `n` and `\161` is `q`.

CHANGES
schmonz: Forgot to mention, by way of +1: I use `config-fast` in pkgsrc too.

CHANGES
```diff
@@ -1,3 +1,7 @@
+20190710 code: add instchown to set ownership and mode on installed files.
+20190709 code: add instpackage to install without root access.
+20190714 code: removed idedit, install-big.  var-qmail packages are no more.
```
schmonz: Sort by date?

alanpost: No, but the reason bears explaining.  There are two dates here, one explicit and one implicit.  The explicit date is when the change was made, the implicit date is the order of the lines in the file, which is the order the commits went in to master.

This file is going to create merge conflicts on rebase, frequently (It's the nature of ChangeLog files and there are format changes we could make that then provide tooling to make it easier to manage, but that's a standalone PR and discussion), because the CHANGES file changes with every substantial commit.

To make it easier to merge, as well as coherent to follow by reading in order, I've kept these lines in commit order, even when the date the change was made is thereby out of order.  By doing this, every change that modifies CHANGES places it's entry at the top of that file, and resolving conflicts is a matter of putting them in commit order, usually by taking non-context lines from both changes in the order presented in a merge conflict.  (Only possible if commit order of CHANGES is maintained.)

To get the implicit date (commit order) and the explicit date (as stated in CHANGES) to match we'd need to invent or triangulate on a date that is in-order for both cases--which is naturally the date we rebase on to master.

Enforcing that will likely (as it would here) result in a last-minute entry to CHANGES across each commit in the series, just to "falsify" the explicit date by moving it forward in time so the whole series has a single date, right before we rebase on to master.  Possible.  Burdensome.

This will result in more merge conflicts than we already create by maintaining a ChangeLog.  Instead of doing that we should use tooling and file formats designed to solve this problem, and leave these lines in-order qua commit history but out-of-order qua date-of-commit until and unless we've done so.  (A quick read of pages return from the web search "changelog merge conflict" will give you a reasonable overview of the problem.)

alanpost:
@DerDakon, @mbhangui,

PR #15 here is the penultimate commit for notqmail-1.07.  It needs one more approval.  This PR has gone through a number of revisions, and I may have lost track of work you're expecting from me.  If you're waiting on any changes to this PR from me, would you please restate them?

If you've still due diligence, testing, or other work to do prior to approval, I completely understand--my message here is to forestall the possibility of deadlock, not to circumvent that necessary work.

DerDakon:
I just lack time to properly look at that, so it's all me that blocks this.

auto_users.h
```diff
@@ -0,0 +1,16 @@
+#ifndef AUTO_USERS_H
+#define AUTO_USERS_H
+
+extern char auto_usera[];
```
DerDakon: all of them should be const.

alanpost: const does not appear in qmail-1.03.  I see we accidentally got one in the DESTDIR patch, and I see netqmail-1.04 added one in sendmail.c, but adding them, like adding ANSI function signatures, exceeds our mandate for the 1.07 release.

We should, for certain, add const systemically in the codebase--it's just not for this release.

alanpost:
> * for instpackage it should be enough to just pass -1, as chown will not change a uid or gid of a file if that argument is -1.

Good catch, I've pushed a new change that fails to look up the uid/gid in instpackage, using -1 instead.

> new code should really not introduce K&R-style functions. Remember mkdir_p(), which was fine when it had a C89 function signature.

I agree that we should transform the codebase to use C89 function signatures.  It is reasonable that, as part of that, new code not introduce K&R signatures.  It is also reasonable that we convert compilation units or the whole codebase at once, rather than when code is touched.  Regardless, converting to C89 signatures exceeds the mandate of this release, so I did not do it.

I also didn't use uid_t/gid_t in this patch, for all the same reason: it exceeds the mandate of the first release, and the whole thing are a repeating series of incremental changes.

instpackage.c
```diff
@@ -0,0 +1,21 @@
+#include "auto_uids.h"
+
+int auto_uida;
```
DerDakon: why not initialize them right here?

DerDakon: ok, forget it, you need the function anyway…

schmonz:
IMO, the only review this PR really needs is security review. We (== @alanpost) have already iterated a whole bunch of times on design, so there's consensus there already. If there are style choices we want to change later, we can. As soon as we can convince ourselves these changes are _safe_, we can get cracking on our first release. I'm very eager to do that, not at the expense of security, but in all other respects ASAP. :-)

instuidgid.c
```diff
@@ -0,0 +1,22 @@
+#include "auto_uids.h"
+#include "auto_users.h"
+
```
DerDakon: #include "uidgid.h"

alanpost: Done.

qmail-queue.c
```diff
@@ -14,6 +14,7 @@
 #include "extra.h"
 #include "auto_qmail.h"
 #include "auto_uids.h"
+#include "auto_users.h"
 #include "date822fmt.h"
 #include "fmtqfn.h"
```
DerDakon: #include "uidgid.h"

qmail-start.c
```diff
@@ -3,6 +3,7 @@
 #include "exit.h"
 #include "fork.h"
 #include "auto_uids.h"
+#include "auto_users.h"
```
DerDakon: #include "uidgid.h"

qmail-queue.c
alanpost: Great catch, fixed.

qmail-start.c
alanpost: Fixed here too, including modifying Makefile to account for the missing uidgid.h.

PR #16, opened 2019-07-09
Mac OS X: explicit initialization due to linker error, rename files to avoid conflict on case-insensitive filesystem.

alanpost:
A couple patches for Mac OS X support.


alanpost:
@schmonz, the remaining work to build on Mac OS X is defining BIND_8_COMPAT=1 during compile so the arpa/nameser.h header included in dns.c includes the necessary declarations.  How would you like to see that done?  I'm not sure it's appropriate to always define this variable (via conf-cc), as the conf-cc file may be modified before the build.

The remaining options are to always #define this variable in dns.c (or always include arpa/namser_compat.h) or to create a build-time test to define it only when necessary.  Thoughts?

Would you mind building on Mac OS X without BIND_8_COMPAT and reporting here the error message?  Like the strerr patch in this series, I'd like to include that in the commit message.

DerDakon:
I wonder if we could just do something like this in dns.c:

```c
#ifdef __mac__ /* whatever is correct here */
#define BIND_8_COMPAT 1
#endif
#include <arpa/nameser.h>
```

schmonz:
Compiling without `-DBIND_8_COMPAT=1`, I get this:

```
./compile dns.c
dns.c:30:18: error: use of undeclared identifier 'MAXDNAME'
static char name[MAXDNAME];
                 ^
dns.c:49:45: error: use of undeclared identifier 'PACKETSZ'
  if (response.buf = (unsigned char *)alloc(PACKETSZ+1))
                                            ^
dns.c:50:21: error: use of undeclared identifier 'PACKETSZ'
   responsebuflen = PACKETSZ+1;
                    ^
dns.c:51:3: warning: add explicit braces to avoid dangling else
      [-Wdangling-else]
  else return DNS_MEM;
  ^
dns.c:53:30: error: use of undeclared identifier 'C_IN'
 responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
                             ^
dns.c:55:37: error: expected expression
     (responselen > 0 && (((HEADER *)response.buf)->tc)))
                                    ^
dns.c:55:29: error: use of undeclared identifier 'HEADER'
     (responselen > 0 && (((HEADER *)response.buf)->tc)))
                            ^
dns.c:60:5: warning: add explicit braces to avoid dangling else
      [-Wdangling-else]
    else return DNS_MEM;
    ^
dns.c:63:33: error: use of undeclared identifier 'C_IN'
    responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
                                ^
dns.c:73:38: error: use of undeclared identifier 'HEADER'
 responsepos = response.buf + sizeof(HEADER);
                                     ^
dns.c:74:22: error: expected expression
 n = ntohs(((HEADER *)response.buf)->qdcount);
                     ^
dns.c:74:14: error: use of undeclared identifier 'HEADER'
 n = ntohs(((HEADER *)response.buf)->qdcount);
             ^
dns.c:74:22: error: expected expression
 n = ntohs(((HEADER *)response.buf)->qdcount);
                     ^
dns.c:74:14: error: use of undeclared identifier 'HEADER'
 n = ntohs(((HEADER *)response.buf)->qdcount);
             ^
dns.c:74:22: error: expected expression
 n = ntohs(((HEADER *)response.buf)->qdcount);
                     ^
dns.c:74:14: error: use of undeclared identifier 'HEADER'
 n = ntohs(((HEADER *)response.buf)->qdcount);
             ^
dns.c:74:22: error: expected expression
 n = ntohs(((HEADER *)response.buf)->qdcount);
                     ^
dns.c:74:14: error: use of undeclared identifier 'HEADER'
 n = ntohs(((HEADER *)response.buf)->qdcount);
             ^
dns.c:77:60: error: use of undeclared identifier 'MAXDNAME'
   i = dn_expand(response.buf,responseend,responsepos,name,MAXDNAME);
                                                           ^
dns.c:81:12: error: use of undeclared identifier 'QFIXEDSZ'
   if (i < QFIXEDSZ) return DNS_SOFT;
           ^
dns.c:82:19: error: use of undeclared identifier 'QFIXEDSZ'
   responsepos += QFIXEDSZ;
                  ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
2 warnings and 20 errors generated.
```

alanpost:
I think it's safe to update the header `arpa/nameser.h` to `arpa/nameser_compat.h`

The `_compat.h` header is available on Mac OS X, CentOS, Ubuntu, and DragonFlyBSD.  I suspect it's available on FreeBSD, NetBSD, and OpenBSD as well.

DerDakon:
OpenBSD does not have nameser_compat.h, NetBSD has it.

strerr_sys.c
```diff
@@ -1,7 +1,7 @@
 #include "error.h"
 #include "strerr.h"
 
-struct strerr strerr_sys;
+struct strerr strerr_sys = {0,0,0,0};
```
DerDakon: Please also add a comment here and say something like "workaround for MacOS linker bug" so one does not need to dig project history to see what this is for.

alanpost: Reasonable, done.

alanpost:
The commit to support a "rename" in the install is my least favorite part of this series, so I'm happy to see an alternative.

alanpost:
I've rewritten commit 0170f04 to define BIND_8_COMPAT prior to including arpa/nameser.h, rather than always including arpa/nameser_compat.h, for compatibility with OpenBSD and other systems that do not have arpa/nameser_compat.h.

I believe this corrects the build problem on Mac OS X, and does so without a runtime feature test.  Please have a look--if this suits (a similar mechanism is used in `ipme.c`) we can avoid the extra runtime feature test.

I'm happy to write said test if anyone has concerns about the above approach, just voice them.

alanpost:
PR #36 is a large change, and I think it should go in 1.08.  To get 1.07 branched, we need PR #16 (Mac OS X port) and PR #15 (runtime uid/gid and package support).  Everything else has gone in, and then some.

I experience us to be on a treadmill of adding features and bugfixes that exceed our mandate for 1.07, and I'm not presently aware of any limit to what any one of us, and therefor all of us in total, want in 1.07.  Bringing in correctness and other bugfixes in to 1.07 is causing other bugfixes to come in behind them, without apparent limit.

It's an old codebase--there is too much work here for a single release.  Further, a release consists of features, the set of which we've already agreed upon and the work remaining being PR #16 and PR #15 as stated.  Once the features are in master, we can branch notqmail-1.07 and work through all outstanding bugfix PRs along with the more aggressive changes we're intentionally waiting for after 1.07 to do.

I'm willing to rename just INSTALL and SENDMAIL to .txt in-tree, removing the rename feature, if that would move this PR along.  Failing that, I would rather release 1.07 with the install "rename" support, and then pull it out again in the 1.08 branch, optionally as part of PR #36 should that be approved.  I'm also willing to rename these files to .txt for the 1.07 release and then rename them to .md for the 1.08 release, deferring all discussion of whether to use markdown in favor of text.

Any of those changes are more modest than PR #36.  That PR is a large enough change, as it includes a change in policy (change in file type), that it is in effect pulling in other work as a side effect.

I'd like to expediently get us to a cut branch by removing whatever we have to, including pieces of PR #16 and PR #15, sufficient to meet our feature roadmap.

schmonz:
You're right to highlight our intention for releases to be small, with just enough change to be worthwhile for users to update. Especially so for 1.07, our first time cutting a release and our first chance to entice the qmail and netqmail user base to come with us. It's hard to identify where a line can be drawn, and it's hard to push for some things to be deferred. Thank you for doing both. I agree with both.

Assuming #36's approach is uncontroversial, let's:

1) Defer #36 to 1.08
2) Keep the original "rename `INSTALL` and `SENDMAIL`" idea here in #16 
3) Implement it in the style of #36: rename those two files with `.md` extensions (both in tree and as installed)

instcheck.c
```diff
@@ -88,6 +88,18 @@ int mode;
   perm(".../",subdir,"/",file,S_IFREG,uid,gid,mode);
 }
 
+void v(home,subdir,to,from,uid,gid,mode)
```
schmonz: Sorry for planting a wrong tree for you to bark up. This is well done, but let's not use `v()` after all -- just `mv INSTALL INSTALL.md && mv SENDMAIL SENDMAIL.md` and adjust `hier.c` and `install-big.c` (while we still have it) to match.

dns.c
```diff
@@ -2,6 +2,9 @@
 #include <netdb.h>
 #include <sys/types.h>
 #include <netinet/in.h>
+#ifndef BIND_8_COMPAT
+#define BIND_8_COMPAT /* if Bind 9, Bind 8 compatibility */
```
schmonz: Simple, self-contained, can't think of any downsides. Obvious in retrospect, which is the hallmark of a hard-earned good choice.

DerDakon: Don't forget a comment here to point fingers on MacOS.

instcheck.c
alanpost: I appreciate your acknowledging initiating an approach that turned out to be a dead-end.  I'd rather two commits, one of which is good, than one commit which isn't.  No harm done, trying it left us a better result.  The rename feature is now pulled in favor of absolutely renaming these files.

dns.c
alanpost: I've added a "Mac OS X:" prefix to the comment, I hope that satisfies.

PR #18, opened 2019-07-11
qmail-remote: avoid recoding CRLF to CRCRLF

DerDakon:
If qmail-remote get's a '\r' (CR, carriage return) in it's input stream (which means that a mail injected into qmail-queue contained a '\r') it will send this unmasked to the network. This patch works this way: if there is a single '\r' somewhere in the input stream it is handled as a line break, so it is send as CRLF to the network. If there is a CRLF sequence in the input stream (normally there should only be bare LFs) it will not double the CR (like qmail-remote normally does) but send this CRLF sequence as is.

### DIFF
--- a/qmail-remote.c
+++ b/qmail-remote.c
@@ -200,6 +200,16 @@ void blast()
     if (ch == '.')
       substdio_put(&smtpto,".",1);
     while (ch != '\n') {
+      if (ch == '\r') {
+        r = substdio_get(&ssin, &ch, 1);
+        if (r == 0)
+          break;
+        if (r == -1) temp_read();
+        if (ch != '\n') {
+          substdio_put(&smtpto, "\r\n", 2);
+        } else
+          break;
+      }
       substdio_put(&smtpto,&ch,1);
       r = substdio_get(&ssin,&ch,1);
       if (r == 0) perm_partialline();
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,4 @@
+- 20200803 avoid sending CRCRLF in qmail-remote if input contains CRLF
 - 20200614 remove maildirwatch
 - 20200614 stop rewriting RCPT TO: domains when they are a CNAME (RFC 5321 5.1)
 - 20200523 doc: give text files .md extensions.
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -1,2 +1,4 @@
 *.o
+unittest_blast
 unittest_stralloc
+blast.c
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -9,10 +9,10 @@ default: it
 
 .PHONY: clean default it test
 
-TESTBINS = unittest_stralloc
+TESTBINS = unittest_stralloc unittest_blast
 
 clean:
-	rm -f $(TESTBINS) *.o
+	rm -f $(TESTBINS) *.o blast.c
 
 it: $(TESTBINS)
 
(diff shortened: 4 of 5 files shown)


mbhangui:
The treatment for bare LF is fine but I think better to leave bare CR unmasked. I think, this will break DKIM and the signature computed with the bare CR will no longer match when the mail body has the bare CR corrected.

DerDakon:
It's a bit of a grey area, IIRC the SMTP RfCs do no not allow either CR or LF to show up without the other. It's just garbage in, garbage out.

schmonz:
@mbhangui, you're doing DKIM in the wild, right? I'd love to know whether this change breaks anything so we can weigh benefits against costs with our eyes open.

mbhangui:
> @mbhangui, you're doing DKIM in the wild, right? I'd love to know whether this change breaks anything so we can weigh benefits against costs with our eyes open.
In real life I haven't encountered mails with bare CR. It could be possible if a mail is being injected and the body contains bare CR. If we do dkim signing before it gets passed to qmail-remote, the DKIM signature computed on the recipient host will not match. So the only case where this will break is

1. If we are injecting emails on a system using qmail-inject and the body contains bare CR.
2. We do DKIM Signing before the mail gets into the queue (e.g. a DKIM qmail-queue replacement) or if we do DKIM signing just before the mail content gets passed to qmail-remote (e.g. using QMAILREMOTE feature).

So we are talking of a case where we have improper mail content (with bare CR) that is passed to qmail-inject.

For all valid, proper mail content, this issue will not happen.

mbhangui:
If we do DKIM signing and we convert the bare CR to CRLF in the DKIM signing program, then this issue will not happen. So when we implement DKIM signing, we should have the same code in the dkim signing process.

schmonz:
I'm tentatively convinced this is worth the risk. I'd be more convinced if we knew of real-world cases where a message with bare CRs had gone out on the wire (not involving DKIM at all) and other SMTP servers rejected or dropped or garbled it.

Would like to hear at least one more person's opinion on this before we merge.

DerDakon:
I have seen this happening when programs (mostly PHP scripts) injects mails with CRLF line endings instead of just LF line endings. You can try yourself doing

```bash
echo -e 'Subject: CRLFTest\n\ncr\r\nlf\n' | qmail-inject -a remote@mail.address
```

And then use wireshark or tcpdump and see that CRCRLF sequences go over the wire, which is illegal.

DerDakon:
Sorry, accidentially deleted wrong branch.

qmail-remote.c
```diff
@@ -200,6 +200,16 @@ void blast()
     if (ch == '.')
       substdio_put(&smtpto,".",1);
     while (ch != '\n') {
+      if (ch == '\r') {
+	r = substdio_get(&ssin, &ch, 1);
+	if (r == 0)
+		break;
```
schmonz, line 206: Looks like tab instead of 2-space indent?

qmail-remote.c
```diff
@@ -200,6 +200,16 @@ void blast()
...
+	if (r == -1) temp_read();
+	if (ch != '\n') {
+		substdio_put(&smtpto, "\r\n", 2);
+	} else
```
schmonz, line 210: Missing a `{`? If not, this looks odd.

DerDakon: It's in line 203.

schmonz (approved):
Thank you for putting in the effort to prepare these tests, and for your patience waiting for my review.

PR #21, opened 2019-07-12
TravisCI: test external patch(es)

DerDakon:
First step towards real regression testing so we don't break common external stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 dist: xenial
 language: c
+env:
+  matrix:
+    - BUILD_NAME="default" QMAIL_PATH=""
+    - BUILD_NAME="TLS" QMAIL_PATH="http://inoa.net/qmail-tls/netqmail-1.06-tls-20190517.patch"
 compiler:
  - gcc
  - clang
@@ -13,6 +17,7 @@ script:
  - sudo useradd -u 204 -g 201 -d /var/qmail qmailq
  - sudo useradd -u 205 -g 201 -d /var/qmail qmailr
  - sudo useradd -u 206 -g 201 -d /var/qmail qmails
- - echo "${CC} -O2 -Wall" > conf-cc
+ - if [ -n "${QMAIL_PATH}" ]; then wget -O extra.patch "${QMAIL_PATH}" && patch -p1 -i extra.patch; fi
+ - sed "s/^cc /${CC} -Wall /" -i conf-cc
  - make
  - sudo make setup check


alanpost:
The idea presented here, testing upstream netqmail patches, is valuable.  The implementation presented in this PR is untenable.

* patches are downloaded from external websites.  Those links may change, go dead, or be temporarily offline.  In addition, in this case, it is an http links (not https), so we also expose the CI to man-in-the-middle attacks.
* some patches are no longer available in downloadable form over the web.
* for each BUILD_NAME, only a single patch is supported.  Some patches do not apply cleanly to netqmail, as the rely on other published patches.  (Some patches don't apply cleanly for other reasons, as well.)
* every patch is assumed to apply with -p1.  In truth some patches need -p0, others -p2, or can be applied with tools like git-am.
* Any conflicting source code change with an upstream patch will cause a CI failure.  This could even include attempts to fix bugs in upstream patches, which upon resolution would otherwise pass CI testing.

Do not use the CI system for branch merging.  Do it in the repository.

I have created the following netqmail branches, corresponding to a variety of published patches.  Including netqmail-1.06-tls-20190517.patch used as the example here:

* [netqmail-smtpd-logging](https://github.com/notqmail/notqmail/commits/netqmail-smtpd-logging)
* [netqmail-badmailfrom-wildcard](https://github.com/notqmail/notqmail/commits/netqmail-badmailfrom-wildcard)
* [netqmail-badmailfrom-x-relayclient](https://github.com/notqmail/notqmail/commits/netqmail-badmailfrom-x-relayclient)
* [netqmail-big-concurrency](https://github.com/notqmail/notqmail/commits/netqmail-big-concurrency)
* [netqmail-big-todo](https://github.com/notqmail/notqmail/commits/netqmail-big-todo)
* [netqmail-dns-any-to-cname](https://github.com/notqmail/notqmail/commits/netqmail-dns-any-to-cname)
* [netqmail-dns-oversize](https://github.com/notqmail/notqmail/commits/netqmail-dns-oversize)
* [netqmail-ext-todo](https://github.com/notqmail/notqmail/commits/netqmail-ext-todo)
* [netqmail-smtp-auth](https://github.com/notqmail/notqmail/commits/netqmail-smtp-auth)
* [netqmail-smtp-tls](https://github.com/notqmail/notqmail/commits/netqmail-smtp-tls)
* [netqmail-smtpd-logging](https://github.com/notqmail/notqmail/commits/netqmail-smtpd-logging)
* [netqmail-smtpd-spf](https://github.com/notqmail/notqmail/commits/netqmail-smtpd-spf)

It is rare for a netqmail patch to not need some modification to be worth merging with other patches or used in a release.  A comprehensive but incomplete checklist of QA due diligence:

* remove trailing whitespace.
* add any new files to FILES.
** merge or remove any spurious files.
* if necessary, add any new files to hier.c/install-big.c doc/
* restore Makefile to alphabetical order.
* sync the Makefile shar target with FILES.
* add any new targets to TARGETS.
* remove any ifdef/define lines.  The branch contains the feature.
* verify Makefile includes any new header dependencies in .o targets.
* add any missing entries to CHANGES.
* update or add any missing manpages or other documentation.
* stylistic cleanup: make the code match the existing codebase.
* commit any necessary bugfixes.

You'll find various examples of the above in the patch branches noted here.  I first pushed the as-published patch and then performed follow-up commits for any or all of the above.

Additionally, as an example of why the branch approach is suitable, here is a notqmail branch of the same netqmail branch with additional local modifications (which exceed correctness fixes against the patch in question):

* [notqmail-smtpd-logging](https://github.com/notqmail/notqmail/commits/notqmail-smtpd-logging)

Allowing us to take netqmail work, modify it or merge it with other patches, and test that, rather than requiring the impossible situation of a notqmail branch cleanly applying a complicity of netqmail patches.

Please create netqmail branches for published patches against the netqmail codebase.

DerDakon:
I think collecting tons of branches in the repo is not a good idea, as one can easily start losing the overview. I have thought about a special testing-repo, where we collect patches, too. They could come with their own apply script then, doing whatever is needed.

The limitation to p1 would be easily solvable even in the current code, as last resort one could simply add a switch-case-statement for every build type.

alanpost:
> I think collecting tons of branches in the repo is not a good idea, as one can easily start losing the overview.

Use branch namespaces, as I've done here (netqmail- and notqmail-), as well as feature branches which merge individual patch branches together (I propose netqmail+ and notqmail+).

DerDakon:
This is as confusing as anything else, especially given the similarity between net and not.

PR #22, opened 2019-07-14
Remove precompiled var-qmail package support

alanpost:
There are two commits here: one to remove the shar rule from Makefile, one to remove idedit and install-big.

The shar target restates the contents of FILES as dependencies on the shar target, a shell archive.  Shell archives are no longer used for distributing software.  While the command could be updated to use an alternative distribution format, like tape archive (tar), that the rule must restate the contents of FILES means any modification to FILES must also be restated here, and in a way that is more likely to cause a merge conflict.

idedit is intended for use by qmail packagers to change the auto_uid* and auto_gid* values in compiled the qmail binaries to values local to the installed system.

We will look up these uid/gid values at runtime, making this program no longer necessary.  (PR #15)

install-big will be replaced by instpackage and instchown, which use install.c/hier.c and do not require restating the runtime file hierarchy.  (PR #15)


DerDakon:
I totally agree on shar. Removing the idedit should IMHO happen in the PR that actually introduces the runtime id lookup.

alanpost:
The shar change was merged to master in PR #27.  The remaining change here is part of pr #15, and on the advice of @DerDakon that change will stay in that PR.

PR #24, opened 2019-07-15
Fix build on utmpx-only systems, such as FreeBSD.

schmonz:
FreeBSD removed <utmp.h> almost a decade ago, leaving only <utmpx.h>:
https://github.com/freebsd/freebsd/commit/cb38fdae7c004eec078edf577e1819ce9760aecb

This is my qbiff-utmpx patch from <https://schmonz.com/qmail/qbiffutmpx/>.
With it, on a system...

- with <utmpx.h> but no <utmp.h>: build proceeds past qbiff.c

- with <utmpx.h> and <utmp.h>: build chooses <utmpx.h>

- with no <utmpx.h>: no change in behavior

Supports goals: being easily packaged by OS integrators
Risks: may run qbiff incorrectly on a system with broken utmpx implementation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1073,7 +1073,7 @@ qbiff.1
 
 qbiff.o: \
 compile qbiff.c readwrite.h stralloc.h gen_alloc.h substdio.h subfd.h \
-substdio.h open.h byte.h str.h headerbody.h hfield.h env.h exit.h
+substdio.h open.h byte.h str.h headerbody.h hfield.h env.h exit.h qtmp.h
 	./compile qbiff.c
 
 qmail-clean: \
@@ -1674,6 +1674,12 @@ compile qsutil.c stralloc.h gen_alloc.h readwrite.h substdio.h \
 qsutil.h
 	./compile qsutil.c
 
+qtmp.h: \
+tryutmpx.c compile load qtmp.h1 qtmp.h2
+	( ( ./compile tryutmpx.c && ./load tryutmpx ) >/dev/null 2>&1 \
+	&& cat qtmp.h2 || cat qtmp.h1 ) > qtmp.h
+	rm -f tryutmpx.o tryutmpx
+
 quote.o: \
 compile quote.c stralloc.h gen_alloc.h str.h quote.h
 	./compile quote.c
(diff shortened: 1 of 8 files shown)


DerDakon:
You can merge several code paths if you just do this for the "old" path:

```c
struct utmp utm;
struct utmp *ut = &utm;
```

qbiff.c
```diff
@@ -1,13 +1,6 @@
...
 #include <sys/stat.h>
-#include <utmp.h>
-#ifndef UTMP_FILE
-#ifdef _PATH_UTMP
-#define UTMP_FILE _PATH_UTMP
-#else
-#define UTMP_FILE "/etc/utmp"
-#endif
-#endif
+#include "utmp.h"
```
DerDakon: Please do not name this like the global file. While basically all compilers on Un*x systems handle local paths before globals ones when using "", they are not required to (and e.g. Visual Studio does not do this). Please rename it to something sensible and use that name here.

schmonz:
Thanks, @DerDakon, that idea simplified the patch a lot.

utmp.h
```diff
@@ -0,0 +1,34 @@
+#include "hasutmpx.h"
```
DerDakon: This file only works because it has no include guard (see below).

utmp.h
```diff
@@ -0,0 +1,34 @@
...
+#include <utmpx.h>
+#define UTMP_INIT struct utmpx *ut
+#define UTMP_USER ut_user
+#define UTMP_OPEN
+#define UTMP_READ_MORE (ut = getutxent())
+#define UTMP_TYPE_MATCHES (ut->ut_type == USER_PROCESS)
+
+#else
+
+#include <utmp.h>
```
DerDakon: include utmp.h from utmp.h, this just begs for trouble.

Makefile
```diff
@@ -685,6 +685,13 @@ chkshsgr warn-shsgr tryshsgr.c compile load
 	hasshsgr.h
 	rm -f tryshsgr.o tryshsgr
 
+hasutmpx.h: \
+tryutmpx.c compile load
+	( ( ./compile tryutmpx.c && ./load tryutmpx ) >/dev/null \
+	2>&1 \
+	&& echo \#define HASUTMPX 1 || exit 0 ) > hasutmpx.h
```
DerDakon: I would still solve it as uint32.h is created currently: have 2 headers in the tree (*.h1 and *.h2) and simply copy the proper one to *.h, this solves the whole #ifdef code.

schmonz: Yes, that's better too. Have done.

qtmp.h1
```diff
@@ -0,0 +1,19 @@
+#include <utmp.h>
```
DerDakon: These are still missing header guards.

DerDakon:
For later reference: this is incomplete, see #57.

PR #25, opened 2019-07-15
symlink TARGETS to .gitignore.

alanpost:
This is an alternative to PR #13, which uses a symlink instead of copying the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+TARGETS
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ SHELL=/bin/sh
 
 default: it
 
+.PHONY: check clean default it man
+
 addresses.0: \
 addresses.5
 	nroff -man addresses.5 > addresses.0
--- a/TARGETS
+++ b/TARGETS
@@ -320,7 +320,6 @@ binm2
 binm2+df
 binm3
 binm3+df
-it
 qmail-local.0
 qmail-lspawn.0
 qmail-getpw.8
@@ -382,7 +381,5 @@ mbox.0
 addresses.0
 envelopes.0
 forgeries.0
-man
 setup
-check
 qtmp.h


DerDakon:
It adds things like "setup" and "check" to .gitignore that should never exist as files.

schmonz:
That's arguably a bug in `TARGETS`. If it is (which we don't need to decide now), then `.gitignore` being a symlink would mean fixing it once.

DerDakon:
No, TARGETS lists Makefile targets. And setup and clean are Makefile targets.

alanpost:
> It adds things like "setup" and "check" to .gitignore that should never exist as files.

Let's remove the phony targets from TARGETS.  If necessary, we can add them to the .PHONY target so GNU make will unconditionally run that rule regardless of whether any of those names exist as files on the filesystem.

Would that make the contents of TARGETS suitable to also be the contents of .gitignore?

DerDakon:
Sounds reasonable.

schmonz:
Sounds exactly like what at least [NetBSD's make does with .PHONY](https://wiki.netbsd.org/tutorials/bsd_make/), and I can't remember a time it didn't have the feature. Guessing this has effectively become universal make behavior since 1998.

alanpost:
I have updated this PR to contain two commits: one removing all phony targets from TARGETS while adding a .PHONY special target.  Then symlink TARGETS to .gitignore.

I believe this change to be compatible with a make that doesn't recognize the .PHONY special target, as it will not treat that rule specially, and effectively ignore it.  I did take care to not make it the first rule in the Makefile.

DerDakon:
I hope it's just the Github UI being bad, I can't distinguish if you added a .gitignore file with the contents TARGETS or a symlink.

alanpost:
It is a symlink.

That is an artifact of how git displays symlinks.  It's because when git clones a repository on a filesystem that doesn't support symlinks, it creates a file where the contents are the path of the symlink.  And therefor this is how git displays symlinks in diff view.

Outside of GitHub, if you looked at the same diff, it would include the line `new file mode 120000`, where mode 120000 is the value of `S_IFLNK`.

alanpost:
@mbhangui,

Would you like to "Rebase and merge" PR #25 here?  It appears ready to go to master:

* It has the correct milestone, 1.07.  We are only placing 1.07 milestone PRs on master until 1.07 is branched for release.
* It has been approved by two of us.  Here @DerDakon and @schmonz.  That is the minimum and sufficient number of approvals necessary.
* As the author, I claim it is ready.  The PR is not marked as draft.
* The branch has been rebased to the latest master and so will rebase cleanly.
* It is passing CI checks, a strict requirement for every PR.

If all of the above also meets with your satisfaction, click the "Rebase and merge" button, which you may or may not have to select from the down-arrow menu on the right side of the green button.

Regardless of the above, if this commit leaves you with any questions or concerns, of any size, do not rebase and merge it to master, but instead raise those questions and concerns here--do not take in to account approval from other people on the team when making your own determination for approval or merge.  Instead consider each PR on it's own merit independently--each of us will catch mistakes the rest of us will miss.

I do believe this is ready to rebase, and it is best if one of us not the original author of the PR puts it on master.  Have a look.

schmonz:
Happy to have you here, @mbhangui!

PR #27, opened 2019-07-15
remove shar target from Makefile.

alanpost:
This target restates the contents of FILES as dependencies on the
shar target, a shell archive.  Shell archives are no longer used
for distributing software.  While the command could be updated to
use an alternative distribution format, like tape archive (tar),
that the rule must restate the contents of FILES means any
modification to FILES must also be restated here, and in a way
that is more likely to cause a merge conflict.


DerDakon:
This should also remove FILES. Noone needs it: if you want go do a release, use ```git archive```. If any packager wants to build a source package, deb and rpm will just use the original tarball + patches + spec file. Nothing but trouble will be lost by deleting that file.

DerDakon:
Have you ever tried running it? It has never worked, because it depends e.g. on BIN.setup that has never existed. Kill it with fire.

alanpost:
This commit now also removes the FILES file.

PR #30, opened 2019-07-16
use <stdint.h> to get a really portable 32 bit unsigned type

DerDakon:
This in turn requires a compiler from the 21st century.


schmonz:
As usual, I like the change, and am concerned only about timing.

Without this fix, what's at risk? Are there systems that won't compile, systems that'll exhibit surprising runtime behavior, and/or something else entirely?

With this fix, what's at risk? Normally here I'd be concerned about users being able to keep applying their patches, but this one is pretty self-contained. What else should we consider about when is best to merge this?

DerDakon:
This will not work on systems that do not support C99, as they likely do not have that header. For all other systems using ```unsigned long``` at all was IMHO a bad idea, ```unsigned int``` is really 32 bit everywhere for ages.

PR #32, opened 2019-07-19
remove FILES.

alanpost:
It demonstrably is not up-to-date, and git ls-files is a suitable
replacement.


alanpost:
This commit has been merged with the commit to remove the shar target in PR #27.  I'll withdraw this PR as a result.

PR #34, opened 2019-07-19
remove systype and attendant platform detection.

alanpost:
It's used in one place, make-makelib, to exclude ranlib on
platforms that no longer exist.


DerDakon:
It would be cool if someone could tests this on a recent hpux system, all others are dead.

schmonz:
> It would be cool if someone could tests this on a recent hpux system, all others are dead.

According to this [HP-UX 11i ranlib(1) manpage](https://nixdoc.net/man-pages/HP-UX/man1/ranlib.1.html), it has `ranlib` these days.

Makefile
```diff
@@ -361,8 +361,8 @@ case.h
 	./compile commands.c
 
 compile: \
-make-compile warn-auto.sh systype
-	( cat warn-auto.sh; ./make-compile "`cat systype`" ) > \
```
schmonz: Was `systype` even being used in any way by `./make-compile` and `./make-load`? I can't see that it was.

schmonz: Oh, you explained this in the very first comment here.

PR #36, opened 2019-07-19
Append .md suffix to text docs. Adjust references.

schmonz:
Rationale, in some sort of order:

- Plain text is already valid Markdown (and we can improve it as we go)
- Markdown is already plain text, readable in all the same ways
- GitHub also renders Markdown nicely in web browsers
- Changing installed names for docs is a non-zero compatibility break,
  but we consider it small enough to tolerate:
    1. Programs are unaffected
    2. People will press \<tab\> and find what they're seeking

Supports goals: Being easily packaged by OS integrators
Broaches non-goals: Breaking compatibility (but tolerably so)
Risks: Users won't find needed info; patches to hier.c may get skewed


alanpost:
For files like INSTALL.mds, which you have renamed to INSTALL.mds.md, what do you think about replacing the first (original) period (.) with a dash (-)?  INSTALL.mds.md visually looks like INSTALL.tar.gz, in that you expect this to be an .md .mds file, whereas using a dash more idiomatically states this is the -mds section of the INSTALL notes.

SYSDEPS.md
```diff
@@ -1,4 +1,4 @@
-VERSION
+VERSION.md
```
alanpost: Let's go ahead and merge PR #33 in front of this PR, which would eliminate this change.

install-big.c
```diff
@@ -75,29 +75,29 @@ void hier()
...
+  c(auto_qmail,"doc","REMOVE.binmail.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.local2alias.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.local2ext.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.local2local.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.local2rem.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.local2virt.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.nullclient.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.relaybad.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.relaygood.md",auto_uido,auto_gidq,0644);
+  c(auto_qmail,"doc","PIC.rem2local.md",auto_uido,auto_gidq,0644);
```
alanpost: If we merged PR #22, either separately or as part of PR #15, we could avoid this part of the patch.

SYSDEPS.md
schmonz: Yes, if we're willing to promote that one to 1.07 (I am).

install-big.c
schmonz: #15 is 1.07, but #22 is currently 1.08 (odd). If we promote it to 1.07 (I'd agree), then yes.

schmonz:
Renaming a little more, so we have `INSTALL-foo.md` instead of `INSTALL.foo.md`: I'm open to the idea. If you're reading this, what do you prefer?

SYSDEPS.md
alanpost: I moved PR #33 to 1.08, let's leave this change in this patch.

DerDakon:
I would spare 2 files from this:

 - COPYRIGHT, as that may where e.g. GitHub as well as other tools automatically look for. I guess most will look for copyright* case insensitive, but still.
 - TARGETS: it's just an internal file of the buildsystem, and it's also where the .gitignore file is symlinked to. Making this markdown gives no benefits as it's just a list of files.

INTERNALS.md
```diff
@@ -84,13 +90,13 @@ At that instant the message has been successfully preprocessed.
...
-   NOT DONE: If there have been any delivery attempts, they have all
-             met with temporary failure. Either way, qmail-send should
-             try delivery in the future.
+- DONE: The message was successfully delivered, or the last delivery
+        attempt met with permanent failure. Either way, qmail-send
+        should not attempt further delivery to this address.
+ 
+- NOT DONE: If there have been any delivery attempts, they have all
+            met with temporary failure. Either way, qmail-send should
+            try delivery in the future.
```
josuah, line 99: If we ever get conversion issues, we can fix them as we encounter them as stated in the comments...

PR #37, opened 2019-07-20
fix overflow in alloc.c

mbhangui:
this patch will add unsigned integer for an overflow check in alloc.c
Matthew Dempsky came up with a fix.
See this discussion at https://marc.info/?l=qmail&m=125184448101814&w=2


alanpost:
Building upon @DerDakon's suggestion, when you add work on top of a third-party commit, here making a change that alters Matthew Dempsky's patch, keep a commit with the original, upstream, third-party change and create second and subsequent commits with your changes to it.

For example, the branch [netqmail-ext-todo](https://github.com/notqmail/notqmail/commits/netqmail-ext-todo) has the original ext_todo-20030105.patch and two changes after it.

This lets us keep track of individual contributions while also making a PR as a whole suitable for notqmail.

mbhangui:
> Please check for `__builtin_add_overflow` being available and use that if it is, it may happen that the compiler is doing clever optimizations and remove the entire check otherwise.

I  have added the change. However, i feel that I need to check if the compiler supports  _builtin_add_overflow and put some #ifdef to make it portable. One method could be a try-builtin-overflow.c in Makefile which defined #ifdef HAS_BUILTIN_ADD_OVERFLOW

alanpost:
I agree, you're on the right track.  @schmonz did this for FreeBSD support in PR #24, and I must do something of the same sort in PR #16 before it can be approved.

The builtin overflow feature came in all at once, so the name of the check can be something like builtin_overflow, removing `add`, even if the check itself must certainly check for "add."

DerDakon:
I'm unsure what the original code was about to do, i.e. if it was not simply n + ALIGNMENT - 1. Once I have understood what's going on I'm do a proper review.

DerDakon:
Yes, that part is clear. But even the original DJB code is carefully crafted to do something special, and I need to get _that_ before being able to judge anything else.

mbhangui:
> Yes, that part is clear. But even the original DJB code is carefully crafted to do something special, and I need to get _that_ before being able to judge anything else.

This from the file CHANGES & I believe it was to fix the bus error on Solaris

"19960812 portability problem: under Solaris 2.4 and possibly other systems, the linker does not give generic alignment to an array of 4096 chars. tnx JP. impact: some subset of the programs would (reliably) die with a bus error; in the Solaris case, maildir2mbox. fix: redefine space in alloc.c to be aligned."

also this:

"19960829 change: used double union in alloc.c. tnx ME."

DerDakon:
The requirements for malloc() are to be aligned, I have had enough trouble with that on my Linux Sparc machine. We should fix it for 1.07 just because we can, and entirely trash that for 1.08 and just trust malloc().

DerDakon:
I rebased the branch on current master, but did not do any changes. What the code tries to do is to round up to the next multiple of ALIGNMENT. If n is close enough to 2^32 then it could indeed overflow. I guess a much more simple fix for the whole thing would be to wrap the first 2 lines of code in `´alloc()`` in an ``if (n < avail)```. If that is true the condition can never overflow, and this needs much less code and checks to get it right.

schmonz:
@mbhangui, I've added some comments in review.

For the initial "fix overflow in alloc.c" commit, please
[adjust the commit's Author to `Matthew Dempsky <matthew@dempsky.org>`](https://github.com/notqmail/notqmail/pull/37#issuecomment-513466019)
and adjust the log message to refer specifically to
<https://marc.info/?l=qmail&m=125213850310173&w=2>).

For your subsequent commits, if we don't decide (yet?) to fully trust `malloc()`, then please squash your changes.
Sometimes the history of how a PR got iterated on is worth preserving in the commit logs, but I don't think this is one of those times.
I'd like to see this branch consist of one commit by Matthew Dempsky followed by one commit from you.
Don't forget to rebase your branch locally and force-push to replace what's here.
And don't forget (since it's been a while) that we can help you with the git details if you've paged them out. :-)

@DerDakon, when the branch looks otherwise ready to merge, can you validate runtime behavior on your Sparc in some basic ways?

DerDakon:
I can, but I still think that the solution is overcomplicated and hard to verify. My vote is for a simple ```if``` as described above.

schmonz:
The one you described [here](https://github.com/notqmail/notqmail/pull/37#issuecomment-526860795)?

DerDakon:
Exactly.

DerDakon:
Done with #109 in a different way.

PR #38, opened 2019-07-20
remove incorrect usage of vfork()

DerDakon:
From "man vfork(3)":

>   The vfork() function shall be equivalent to fork(), except that the behavior
>   is undefined if the process created by vfork() either modifies any data other
>   than a variable of type pid_t used to store the return value from vfork(), or
>   returns from the function in which vfork() was called, or calls any other
>   function before successfully calling _exit() or one of the exec family of
>   functions.

All of these instances do some filedescriptor modifications before calling execve(). Since these were the only users of vfork() the tryvfork detection can entirely go away.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,5 @@
-20190805 version: notqmail 1.07.
+201908xx version: notqmail 1.07.
+20190819 remove incorrect usage of vfork()
 20190710 code: add instchown to set ownership and mode on installed files.
 20190709 code: add instpackage to install without root access.
 20190714 code: removed idedit, install-big. var-qmail packages are no more.
--- a/Makefile
+++ b/Makefile
@@ -632,13 +632,6 @@ forgeries.0: \
 forgeries.7
 	nroff -man forgeries.7 > forgeries.0
 
-fork.h: \
-compile load tryvfork.c fork.h1 fork.h2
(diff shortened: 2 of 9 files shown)


alanpost:
This PR is perfect.  It fixes a correctness issue.  In doing so, it removes a try* build-time program.  Finally, it's size is quite modest, managing to avoid easy nearby work that is otherwise tempting, staying focused on one thing.

Fantastic work.

schmonz:
For an example of a bug this will fix, see the [dietlibc TODO](https://github.com/ensc/dietlibc/blob/4040a0a024fe9eac7a226df2f07f89911e09f4ab/TODO#L12).

schmonz:
For another example of a bug this will fix, here's me following `TEST.deliver` on OS X 10.14.6:

```
$ echo to:     schmonz | /var/notqmail/bin/qmail-inject     # works
$ echo to: nonexistent | /var/notqmail/bin/qmail-inject     # logs "warning: trouble injecting bounce message, will try later"
```

@alanpost suggested this PR might help us get a more meaningful system-call trace. I switched to this branch, ran `make setup check`, restarted `/var/notqmail/rc` (the one from LWQ), and the bounce message was injected and delivered.

Trying the initial sequence again, both the successful delivery and the bounce delivery worked immediately without warnings.

PR #42, opened 2019-07-22
Update docs: we're not netqmail, we're notqmail.

schmonz:
We don't know the 1.07 date yet. When we do, assuming this PR is ready to go and everything else is in order, we can bump the date and merge this.


alanpost:
As we already need a late-to-apply date fix here, let's merge this after PR #15 and, on the assumption that PR is in fact approved and that it or the release process broadly doesn't scare up additional PRs we'll branch notqmail-1.07 from it.

PR #43, opened 2019-07-23
cleanup: remove readwrite.h, use unistd.h instead.

alanpost:
The header readwrite.h (re)declares read(2) and write(2).  We should use the system-provided header for these declarations, unistd.h.

This patch systematically removes readwrite.h and includes unistd.h instead.


DerDakon:
Both share a common problem: they touch a lot of files, which easily will break external patches that add their own headers. I would suggest that we split this into 2 PRs, and #44 as well: one that modifies readwrite.h to just ```#include <unistd.h>```, and one that entirely get's rid of that header. This way we can merge the first one e.g. in 1.08, adding correctness while still allowing external patches to easily be applied. And the other one could come in a later release when we decided to allow patch breakage.

alanpost:
> I would suggest that we split this into 2 PRs, and #44 as well: one that modifies readwrite.h to just `#include <unistd.h>`, and one that entirely get's rid of that header.

I've done this here, and as a consequence set this PR to the 1.9 release.  I've submitted PR #80 for the beginning of this series, with the minimal change as you suggest, for the 1.08 release.

Issue #7, opened 2019-07-08
Stop redeclaring system functions


DerDakon:
We should really include <stdlib.h> and friends and get rid of all the handwritten forward declarations.

I don't think anyone remotely sane should care about SunOS or equally outdated system compatibility anymore. My rule of thumb would be: if it hasn't seen a decently complete C99 implementation we should not care. These machines should be kept away from the internet as far as possible. Oh, wait, the 80's have called and want their rsh security holes back…

alanpost:
Minimally this would eliminate the long-standing issue regarding errno in the qmail codebase, where `extern int errno` doesn't work on at least Linux due to macros implementing thread-local storage (if my memory is true).

The level of cross compatibility, correctness, and utility of system headers is materially different than it was in the 90's.  I would like to see the standard headers used everywhere in the codebase for their stated purpose.

Retrofitting doesn't even have to involve touching every file: a single header can include all possible and necessary system headers and that file can be precompiled for performance, if necessary.  (My testing indicates precompiled headers in C projects are marginal to no performance difference.)

As such, agreed.

DerDakon:
Touching every file will be no problem I guess, we can probably let IWYU or similar tools do the stupid work for us.

schmonz:
This should be on our roadmap, no question. I'm worried that if we make a change like this now, before we've implemented most of the features people are currently applying patches for, we'll make it unnecessarily costly for would-be notqmail users to apply their patches, and we'll unnecessarily lose those users. (Related to this, earlier today I added a [What happens to my custom patchset?](https://github.com/notqmail/notqmail/wiki#what-happens-to-my-custom-patchset) section to the wiki.)

Maybe I'm wrong and this particular change wouldn't affect any users. Rather than speculate, I'd like to set up an automated regression test: select a baseline set of popular patches, verify that each patch applies on its own, and maybe verify certain combinations too. Then we can run the regression test for a change like this one, and if it's still passing, I have nothing to be concerned about :-)

DerDakon:
This is definitely nothing for the first 2 releases, agreed.

schmonz:
Added the regression-test idea as [#17].

alanpost:
PR #43 and PR #44 draw down, in their own way, the pervasive redeclaration of system functions.

PR #44, opened 2019-07-23
cleanup: remove exit.h, use in unistd.h.

alanpost:
This PR removes exit.h to replace it with, unistd.h.  The same approach was used in PR #43, which removes readwrite.h, so this PR builds on that one.  They touch a lot of the same files.

exit.h only includes system-defined functions, and since _exit is so basic to a process it shows up all over the codebase.  It's a reasonable start for removing redeclaration of system functions (issue #7), as it touches a lot of files and includes a system header that has the bulk of the declarations we no longer want to shadow.

Alas, exit.h removal cannot be done before PR #43, as including unistd.h causes errors including the readwrite.h header.  As a result I'm experimenting here with separate PR for otherwise staged commits.  That was troublesome for me in the 1.07 series, forgive the experimentation.


DerDakon:
The target branch of this PR is not master, which looks odd.

alanpost:
> The target branch of this PR is not master, which looks odd.

As this series of commits is split in to those suitable for 1.08, and those targeted for 1.9, I've set the target branch to master.  I have cherry-picked the 1.08 readwrite.h commit from PR # 80 instead of the 1.9 readwrite.h PR #43 , as this exit.h PR absolutely requires something be done about the declarations in readwrite.h but doesn't require removing that header in favor of unistd.h.

As a result each PR (#43 and #44) now have fewer dependencies on the other.  I've updated the milestone for this PR to the 1.9 release.  I've submitted PR #79 for the beginning of this series, with the minimal change for the 1.08 release.

PR #45, opened 2019-07-23
use uid_t and gid_t in prot.c

DerDakon:
This should properly reflect if this is a 16 or 32 bit type.

This collides with the dynamic lookup patches, I just want to have it public so I can't forget that this is also there.


DerDakon:
I intentionally kept the check, which is indeed something we could argue about.

DerDakon:
I misunderstood the check I think. I had the impression that it would check if those interfaces are working and warn about a broken state, but it just looks for the right type size which we now assume to have. So removing it is fine.

DerDakon:
#72 is closer to success, no need for this anymore.

Issue #2, opened 2019-07-07
Relook at /var/qmail for everthing


mbhangui:
Most linux distribution follow FHS standards. Look at http://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html

We should relook at having binaries in /var/qmail/bin, man pages in /var/qmail/man, etc. These directories also cause warnings/errors when you use a packing build service like openSUSE build service. rpmlint also lists such violation of FHS as error instead of warning. One of our goalis should be to make notqmail instlallation trivial for many class of users, something as easy as

yum install dotqmail
or
apt-get install dotqmail

The change to FHS is not easy. Many of the qmail programs have
   chdir("/var/qmail");
....
    args[0] = "bin/qmail-local";
    args[1] = "--";
    args[2] = x;

If the goal of dotqmail will never be seeing it installed as a standard (like postfix, sendmail, exim), then this is not required. But if it needs to be made poplular with all class of users, then we should take a relook at /var/qmail for everhing. Even something like /opt/qmail will serve better.

schmonz:
Having notqmail available in system-native package repositories is an explicit [goal](https://github.com/notqmail/notqmail/wiki#goals), "Being easily packaged by OS integrators". It sounds like you're seeing a conflict with our explicit [non-goal](https://github.com/notqmail/notqmail/wiki#non-goals), "Breaking compatibility (such as by moving away from `/var/qmail`)". Thank you for raising this.

I have some experience as a package maintainer in [pkgsrc](https://www.pkgsrc.org), where packages need to install files in locations consistent with [BSD hier(7)](https://netbsd.gw.com/cgi-bin/man-cgi?hier+7+NetBSD-current). I've solved the problem in [pkgsrc qmail](http://pkgsrc.se/mail/qmail) like so:

```
$ ls -l /var/pkg/qmail
total 0
lrwxr-xr-x  1 root  wheel  20 Jun 30 20:10 alias -> /etc/qmail/alias
lrwxr-xr-x  1 root  wheel  22 Jun 30 20:10 bin -> /opt/pkg/bin
lrwxr-xr-x  1 root  wheel  44 Jun 30 20:10 boot -> /opt/pkg/share/examples/qmail/boot
lrwxr-xr-x  1 root  wheel  22 Jun 30 20:10 control -> /etc/qmail/control
lrwxr-xr-x  1 root  wheel  34 Jun 30 20:10 doc -> /opt/pkg/share/doc/qmail
[...]

PR #46, opened 2019-07-24
Run alternate qmail-remote by setting QMAILREMOTE.

schmonz:
This is like the QMAILQUEUE patch, but for qmail-remote(8).

With this change, on a system wherein qmail-start's environment...

- has QMAILREMOTE set: it's invoked by qmail-rspawn(8) instead of
  "qmail-remote"

- has QMAILREMOTE unset: no change in behavior

A QMAILREMOTE program must take care to adhere to the qmail-remote(8)
interface. It can wrap the original, concluding its work by running
qmail-remote, or it can be a full replacement.

Supports goals: "Adding interfaces and seams, where needed, to make extensions possible"
Risks: none known

### DIFF
--- a/qmail-rspawn.c
+++ b/qmail-rspawn.c
@@ -8,6 +8,7 @@
 #include "uidgid.h"
 #include "auto_uids.h"
 #include "auto_users.h"
+#include "env.h"
 
 int auto_uidq;
 
@@ -83,6 +84,16 @@ int len;
     }
 }
 
+static char *setup_qrargs()
+{
+ static char *qr;
+ if (qr) return qr;
+ qr = env_get("QMAILREMOTE");
+ if (qr) return qr;
+ qr = "qmail-remote";
+ return qr;
+}
+
 int spawn(fdmess,fdout,s,r,at)
 int fdmess; int fdout;
 char *s; char *r; int at;
@@ -90,7 +101,7 @@ char *s; char *r; int at;
  int f;
  char *(args[5]);
 
- args[0] = "qmail-remote";
+ args[0] = setup_qrargs();
  args[1] = r + at + 1;
  args[2] = s;
  args[3] = r;
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,4 @@
+20190826 code: if $QMAILREMOTE is set, it's invoked instead of qmail-remote.
 20190822 cleanup: remove unused dnsmxip.
 20190822 cleanup: remove unused dnscname.
 20190819 version: notqmail 1.07.
--- a/Makefile
+++ b/Makefile
@@ -1483,11 +1483,11 @@ tcpto.h readwrite.h timeoutconn.h timeoutread.h timeoutwrite.h
 
 qmail-rspawn: \
 load qmail-rspawn.o spawn.o tcpto_clean.o now.o coe.o sig.a open.a \
-seek.a lock.a wait.a fd.a stralloc.a alloc.a substdio.a error.a str.a \
+seek.a lock.a wait.a fd.a stralloc.a alloc.a substdio.a error.a env.a str.a \
 auto_qmail.o uid.o auto_userq.o auto_spawn.o
 	./load qmail-rspawn spawn.o tcpto_clean.o now.o coe.o \
 	sig.a open.a seek.a lock.a wait.a fd.a stralloc.a alloc.a \
-	auto_qmail.o uid.o auto_userq.o substdio.a error.a str.a \
+	auto_qmail.o uid.o auto_userq.o substdio.a error.a env.a str.a \
 	auto_spawn.o
 
 qmail-rspawn.0: \
@@ -1495,7 +1495,7 @@ qmail-rspawn.8
 	nroff -man qmail-rspawn.8 > qmail-rspawn.0
 
 qmail-rspawn.o: \
-compile qmail-rspawn.c fd.h wait.h substdio.h exit.h fork.h error.h \
+compile qmail-rspawn.c fd.h wait.h substdio.h exit.h fork.h error.h env.h \
 tcpto.h
 	./compile qmail-rspawn.c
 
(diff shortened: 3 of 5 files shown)


schmonz:
This one is intended for our "Introduce new programming interfaces for use by extensions" release. The implementation may change depending on how we implement the "FHS-strict" option of #2 (no problem, that's scheduled for 1.08).

qmail-rspawn.c
```diff
@@ -84,7 +95,9 @@ char *s; char *r; int at;
  int f;
  char *(args[5]);
 
- args[0] = "qmail-remote";
+ setup_qrargs();
```
DerDakon: This repeats the same steps every time. Isn't there some way we could hook this into main()?

qmail-rspawn.c
```diff
@@ -84,7 +95,9 @@ char *s; char *r; int at;
...
+
+ args[0] = qrargs[0];
```
DerDakon: setup_qrargs() could just return the value. Then qrargs could be a static variable just in that function.

mbhangui:
I have done something similar in indimail-mta in qmail-lspawn.c and qmail-rspawn.c by setting QMAILLOCAL and QMAILREMOTE. Instead of calling qmail-local  / qmail-remote, I call a program called spawn-filter which can run user defined filters and the output of that gets passed to qmail-local, qmail-remote. @schmonz, take a look at qmail-rspawn.c, qmail-lspawn.c and spawn-filter.c.  I use it mainly for DKIM.
https://github.com/mbhangui/indimail-mta/tree/master/indimail-mta-2.8

qmail-rspawn.c
schmonz: Good idea.

qmail-rspawn.c
schmonz: Agreed. When I wrote the patch, I was trying to make it look as much as possible like [the QMAILQUEUE patch](http://qmailorg.schmonz.com/qmailqueue-patch). I'm not sure why I thought that was important at the time.

schmonz:
@mbhangui, that sounds more generalized than what I've done here, and might be preferable. Would you be willing to submit a PR with it? No hurry whatsoever, this would be for a couple releases after 1.07.

mbhangui:
> @mbhangui, that sounds more generalized than what I've done here, and might be preferable. Would you be willing to submit a PR with it? No hurry whatsoever, this would be for a couple releases after 1.07.

Sure. I will also make the change that @DerDakon suggested and remove all indimail related changes. The change will add a control file named /var/qmail/control/filterargs where one can specify filters
The  format  of  this file is of the form domain:args for both local and remote mails. domain:remote:args for remote mails and domain:local:args for local mails. Here domain refers to the recipient domain. If there are multiple lines which  match  domain, the first line that matches domain will be used. e.g. a line in the control file filterargs

indimail.org:remote:/usr/bin/dk-filter

will cause all emails sent to indimail.org, get filtered through the the program /usr/bin/dk-filter. The program /usr/bin/dk-filter will just take the original email on stdin and print the email on stdout with the DKIM-Signature added.

DerDakon:
If this becomes code like this:

```c
const char *qrprogram()
{
  static const char *qrprog;
  if (qrprog != NULL)
    return qrprog;
 …
}
```

then it does not need to be in main as it will be only on if call per scheduled mail, regardless which binary will finally be called.

mbhangui:
@schmonz I have done the changes. spawn-filter.c and wildmat.c will require a good review. How do I put it here? Do I use this same branch alternate-qmailremote and add changes (this is what I have done now, but I have not committed or pushed the changes)
   or
do I create a new pull request after your changes are merged?

Following will be the changes
1. No change to qmail-rspawn.c. Your change is perfect and it assigns qrargs[0] just once
2. Same change added to qmail-lspawn.c using env variable QMAILLOCAL
3. Modification of man pages qmail-rspawn.8, qmail-lspawn.8 for QMAILREMOTE, QMAILLOCAL
4. Modification of man page for qmail-control (for new control file filterargs)
5. addition of a new program spawn-filter (which can be called by setting QMAILREMOTE or QMAILLOCAL
6. Addition of a C file wildmat.c for wildmat_internal() function
7. modification of Makefile, TARGETS
8. modification of hier.c for installing spawn-filter and spawn-filter man page

qmail-rspawn.c
mbhangui: @schmonz I can see that you have done the change @DerDakon requested. I believe you can hit the 'Resolve conversation' to indicate that you have made the change.

schmonz: You're right, except that I haven't lifted a finger to do anything new with this code yet :-)  Too far in the future, and too curious to see your alternative.

schmonz:
@mbhangui, I think we'll want to be able to look at all our options for how to solve this problem. (We did this for DESTDIR. One PR got merged, and the rest closed.) So please create a new PR on a new branch, with the name not being terribly important because it won't stick around forever. When I see the PR land, I'll reference it from this one. Thanks!

mbhangui:
Should I click on the "Rebase and Merge". The branch has no conflicts.

mbhangui:
> That would work, but I'd like to see some comparative opinions on your #50 before we decide what to merge for 1.08. :-)
The difference are the following
1. #50 can run alternate qmail-remote as well as alternate qmail-local
2. It has an additional binary spawn-filter which can run either qmail-local or qmail-remote before calling the actual qmail-remote / qmail-local
3. Additional complexity be introducing a new binary named spawn-filter

mbhangui:
We can consider #50 after spawn-filter.c is thouroughly reviewed. Merging #46 for now.

qmail-rspawn.c
```diff
@@ -83,14 +84,24 @@ int len;
     }
 }
 
+static char *setup_qrargs()
+{
+ static char *qr;
```
josuah, line 89: is it not required to set `qr = NULL` here ?
I fear `if (qr)` just below has undefined behavior as on the first run, `qr` has no value set.

Using a static variable for caching the result of env_get sounds good though.

DerDakon: static variables are 0 by default.

PR #54, opened 2019-08-05
replace many pobox.com URLs

DerDakon:
While at it replace some FAQ answers touched anyway with the versions from online.


schmonz:
Testing a handful of pobox.com URLs, either they still redirect directly to the right places, or DJB has reorganized his site and left a reference. Based on that, seems not urgent. Haven't read for the other kinds of diffs yet. Could you split into two commits, one for mechanical pobox changes and another for content updates? Maybe the latter will make a stronger argument for merging sooner.

DerDakon:
I deliberately did not split this as this would just force me to update all FAQ items. It's just that when I updated the URL and found by looking at the answer that it was very different from what we had that I just copied the entire paragraph.

PR #56, opened 2019-08-06
Don't install catted manpages

schmonz:
Don't install catted manpages.

FreeBSD 12, at least, no longer includes `nroff(1)` or equivalent in base. Inconveniently, this breaks the build. Conveniently, the need for preformatted "cat pages" has been obviated by modern `man(1)`.


schmonz:
Other affected platforms:
- OpenBSD
- Void Linux

The approach in this PR messes with `hier.c` and risks breaking people's patches. An alternative raised by @alanpost that I quite like for 1.07: one of us builds the `*.0` pages in the `man` target and commits the built catpages. This is much more conservative, and sometime after 1.07 (at our leisure, really) we can think through whether and how to stop providing catpages.

DerDakon:
I thought abou this, especially about the possible collisions when removing the cat man things from install. This would only collide if a patch adds it's own manpage. Does any of the relevant patches really does this? I doubt any does, and even if the fallyout should be easily fixable. So I would vote to take this as it is, and just remove the cruft.

alanpost:
There are two ways in that I do not like this change, and I need to be clear about them as separate issues:

1) I do not like this change for 1.07--as there are multiple alternative, superior resolutions to the build issue you discovered.
2) I do not like this change on it's merit, for 1.07 *or* any later release--with the understanding that the change as such does itself have merit--but that I do not believe this PR fully accounts for the costs it imposes on all platforms by removing cached manpages.

I would like to defer accounting for point 2 here, either moving that discussion to one or more issues and/or putting it aside for the 1.07 release, and focus on point 1.

roff/nroff/groff are not available by default on every platform we test.  Rather than suggesting a change to the code, this suggests a build-time dependency on groff--much like we have a build time dependency on cc (via gcc or clang).  The problem then can be solved by expressing this build-time dependency in the package system.

It is also true, at least prior to this patch, that at runtime roff/nroff/groff is not strictly required--the cached, pre-rendered pages (catpages) can be used by man in lieu of rendering the roff markup to text (console).  On a system that support catpages groff (as nroff) is only a build-time, and not a run-time requirement (at runtime it is optional).  This change makes it a runtime requirement rather than a build time requirement.  I believe that is inferior to the current condition.

It is further true that building documentation is distinct from building software--in the Makefile one is the `it` phony target and the other is the `man` phony target.  groff is only required for the `man` phony target.  As you discovered, however, both targets are required for setup to work, requiring each tested platform to support both our source code build and our documentation build.

I believe this problem is structural and systemic--that it is manageable here via build-time dependency on groff, but that if our documentation was prepared using docbook, or Sphinx, or other frankly larger third-party software--we would find it intolerable to require each tested platform to both support our source code build (`it`) and our documentation build (`man`).

Now, that is not our real problem--our real problem is that we need to express in each package system, including pkgsrc, a build-time dependency on groff.  But we know there is work waiting for us in the documentation.  #36 being our present example, fated in part by #40 which introduced markdown (in the mime-type sense) alongside roff (in the markup language sense) in our packaged documentation.  Such that it is reasonable to ask about removing groff even as a build-time dependency.

As this change does not remove groff as a runtime dependency, and in fact creates one where there was not (a non-optional) one previously, I would rather for the 1.07 release see a change that removes groff as a build-time requirement (making it optional) and preserve groff as an optional runtime requirement (rather than making it non-optional).

The mechanism available to accomplish this is to prepare the documentation on a suitable machine and to carry the essentially platform-independent[1] files in to a build host that is otherwise preparing a binary (software) release.  This can be accomplished by including the prebuilt catpages in a distribution tarball, a git branch, or otherwise bundled alongside the source code.

I propose first that we arrange for the source code tarball to include these files, leaving them out of git.  If that is unsatisfactory, I propose we prepare a git branch with these files, using a commit not intended to be merged back to master, but suitable for users who want to deal with notqmail releases.

On the power of my comment here, which I hope completely accounts for the issue, I'm going to remove the 1.07 milestone and defer all but the discussion of expressing a runtime dependency on groff to after the release

1: the manpages variously do include the value of conf-qmail, conf-split, &c and so are only platform-independent between otherwise identically configured builds.

schmonz:
The proposed changes here rely on the claim that there exist no systems we'd want to care about where precompiled manpages are still needed. It's a strong claim, and at present I don't know how to make it with complete confidence. I do know pkgsrc (which supports quite many platforms) decided years ago to move away from installing precompiled manpages. But that's not exhaustive. So I'm willing to agree that these changes are insufficiently conservative for 1.07, and would need stronger evidence before being considered for some future release.

As a notqmail _packager_, I'm happy to note that there's build-time requirement for *`roff` and let the package system help me get it. As a notqmail _maintainer_, I'd be sad to have our first release require users of several common platforms to go install an old tool that's not otherwise needed on their systems (as evidenced by its having been removed ;-)). I think we're in strong agreement here too.

That the manpages — also including the precompiled ones — encode some build-time configuration values suggests that the documentation needs to get built as part of the regular build (as it has been until now), or else managed separately and somehow tightly coordinated. Else we run the risk of having binary packages where the docs don't match the running code. And for that reason, my possibly simplistic preference is for docs to be part of the regular build. If that winds up presenting too high a cost to users doing their own builds, I'm willing to form a new preference. But I think we can and should defer the larger doc-toolchain problem for after 1.07.

For 1.07, another possibly simplistic preference I hold is, everything that's part of a release should have the same provenance — in our case, git at a specific tag. Anyone, including us, can get exactly everything we shipped by going to the one place. Given this preference, I'm wondering: wouldn't it be sensible to have one of us build the `*.0` pages, add them to git, submit a PR, and merge it?

alanpost:
> That the manpages — also including the precompiled ones — encode some build-time configuration values suggests that the documentation needs to get built as part of the regular build (as it has been until now), or else managed separately and somehow tightly coordinated.

Not quite true, but the reason bears explaining.  Imagine, for the moment, qmail is in FHS world.  You've got qmail-newu in /usr/local/bin/, the queue is sitting in /var/lib/qmail/queue/, and your users directory is /etc/qmail/users/.  You want to read the manpage for qmail-newu, so you load up your web browser and visit https://notqmail/ which, if you'll continue imagining, is hosting a copy of our manpages.

Do you expect these documents to have your local paths?  Absolutely not--they all refer to /var/qmail for reason of history, simplicity, reference, and shorthand.  The paths selected by your local installation may vary--the canonical set of man pages simply cannot accommodate the choices individual package maintainers make, even when those choices are internally consistent for the platform in question.  You simply cope.

If this is intolerable, preparing localized versions of your documentation is as describable as it has ever been.

> I'm wondering: wouldn't it be sensible to have one of us build the *.0 pages, add them to git, submit a PR, and merge it?

I do not find this sensible, except in the narrow circumstance I described: a dedicated branch, sufficient to prevent this commit from landing on master.  If the documentation is changed it will also require regenerating the derived files.  We already demonstrably struggle with performing that kind of work, and I'm not keen to add more of it.  We removed the shar target, and FILES, for example.

We also may find (I have with other documentation tools) that our respective platforms generate trivial but material differences in documentation--like added or missing spaces, differing widths of our terminal, or other version dependent results.  If we check in a master copy, it could be that only one of us can reproduce the result.  It doesn't matter if the cache is out of master, and not needed just because we made a documentation change on master.  Making a branch does not commit us to keeping the cached results in sync with the codebase (master), and the attendant problems I'm outlining here.

Plus, finally, that is just going to be an enormous commit.  It will dominate the total lines of code, which right this moment are actually negative for the 1.07 release--we've removed more lines than we added.  Call me sentimental, I like that we did that.

Really, just install the correct tool at build time (groff).  As a courtesy we can put the cached results on a branch--that does not impose on anyone else and is a service to folks that do not want to install groff.

schmonz:
As usual, I find your reasoning thoughtful and mostly convincing. So now I'm trying on a new opinion. If we already have trouble remembering all the mechanical things we want to have done, doesn't committing precompiled manpages to a non-trunk, non-PR, long-lived branch both (1) add one more mechanical thing we're sort of committing to try to remember and (2) be even harder to remember than the mechanical things that are at least included in our usual build and release processes?

My new opinion for 1.07: let's change nothing in the repo, and document in the install notes that

- There's a non-optional build dependency on "nroff"
- Some systems (FreeBSD, OpenBSD, Void Linux...) no longer include such a tool
- `groff` should do the trick

DerDakon:
Do  we want to add a deprecation note to the 1.08 release notes for them, too? We are going to kill some old cruft right after 1.08, and I think this would be good company for them.

DerDakon:
We should at least define that we will not install them by default anymore, so if one wants them we could optionally keep them. The truth is that most distros just delete them during packaging anyway because they do not need catted manpages.

josuah:
Note that if we keep a non-configurable groff, we drag an autotools dependency, then an m4, then autotools again (dependency loop).

Anyone willing to avoid that would have a `sed -i 's/groff/true/' *`: for instance a router makers proposing an SMTP gateway: none of the cat or manpage are needed onboard as it is unexpected that the router users tinker the mail stack in-situ.

DerDakon:
Can this be closed now that we have merged #116?

PR #57, opened 2019-08-06
Fix the non-compiling utmp case, found on OpenBSD.

schmonz:
By not having build-tested with utmp, I made two easy mistakes:

(1) In file included from qbiff.c:3:
    ./qtmp.h:15:5: error: unknown type name 'substdio'
        substdio ssutmp; \
        ^
    ./qtmp.h:16:43: error: expected ';' after top level declarator
        char bufutmp[sizeof(struct utmp) * 16]
                                              ^
                                              ;

Mistake: missed a trailing backslash on line 14.
Fix: add it.

(2) qbiff.c:81:2: error: use of undeclared identifier 'fdutmp'; did you mean 'ssutmp'?
     UTMP_OPEN;
     ^
    ./qtmp.h:20:33: note: expanded from macro 'UTMP_OPEN'
        substdio_fdbuf(&ssutmp,read,fdutmp,bufutmp,sizeof(bufutmp))

Mistake: tried to get rid of fdutmp by checking open_read() directly,
not realizing fdutmp was still needed on the next line (!).
Fix: bring back fdutmp.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,18 @@
 dist: xenial
 language: c
+env:
+  matrix:
+    - BUILD_NAME="default"
+    - BUILD_NAME="utmp" FORCE_UTMP="1"
 compiler:
  - gcc
  - clang
 script:
  - echo "${CC} -O2 -Wall" > conf-cc
- - make it man
- - env DESTDIR=/tmp/qmail make package
- - sudo make package
+ - if [ -n "${FORCE_UTMP}" ]; then export MAKEFLAGS="${MAKEFLAGS} -o qtmp.h"; cp qtmp.h1 qtmp.h; fi
+ - make ${MAKEFLAGS} it man
+ - env DESTDIR=/tmp/qmail make ${MAKEFLAGS} package
+ - sudo make ${MAKEFLAGS} package
  - sudo groupadd -g 200 nofiles
  - sudo groupadd -g 201 qmail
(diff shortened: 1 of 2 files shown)


schmonz:
Good call, updated the commit message.

alanpost:
For the benefit of readers not in front of an OpenBSD box, will you explain operationally what "fix" you needed because this change was "non-compiling?"  You have access to a lot more information that I would myself like to be privileged to--will you amend the commit message explaining what happened rather than what you did?

DerDakon:
The macro did not properly escape all continuation lines.

alanpost:
> The macro did not properly escape all continuation lines.

Even saying that much is an improvement over the current message, agreed.  It also left out a temporary variable.

.travis.yml
```diff
@@ -1,10 +1,15 @@
...
+env:
+  matrix:
+    - BUILD_NAME="default"
+    - BUILD_NAME="utmp" FORCE_UTMP="1"
 compiler:
  - gcc
  - clang
 script:
  - echo "${CC} -O2 -Wall" > conf-cc
+ - if [ -n "${FORCE_UTMP}" ]; then cp qtmp.h1 qtmp.h; touch -t 203801012359 qtmp.h; fi
```
alanpost: Can we use make -o qtmp.h instead of a touch with a future date?

schmonz: Yes, @DerDakon provided this and it runs correctly green in Travis.

PR #58, opened 2019-08-10
TravisCI: minor improvements

DerDakon:
* avoid indirection for MAKEFLAGS setting
 * enable reports in IRC channel


DerDakon:
I removed the IRC notifications as this will be solved differently.

PR #35, opened 2019-07-19
split conf-qmail in to conf-qmail-bin, conf-qmail-queue, conf-qmail-man, etc.

alanpost:
This is a technology preview.  This PR is not ready to merge.

PR #61, opened 2019-08-12
Extract an "instqueue" program from instpackage.

schmonz:
Packagers will likely want to:

(1) Stage installation without creating a queue in the staging area
    (especially for FHS-aware packages, since the queue is outside the
    usual installation prefix).

(2) Create the queue (without reinstalling any other files) when the
    package is installed.

With this change, instpackage defaults to (1), and does (2) when
INSTQUEUE is provided in its environment.


alanpost:
I think this PR is overfit for the problem it is solving, though I acknowledge the problem is real enough--I've written scripts to recreate the qmail queue--not for this precise reason but for packaging reasons.

If we need to create the queue, let's create an instqueue program rather than defining an environment variable that changes the behavior of instpackage.  If we don't want an instqueue program, add a command-line option to instpackage to control it's behavior and don't use an environment variable.  Either of those technically solves this problem in a more idiomatic way.  Environment variables are for when you don't have access to the command-line of the program, but it is in your calling chain.  Just not true here.

There are eight directories qmail uses: alias, bin, boot, control, doc, man, queue, and users.  What is it that makes queue special here, that does not apply to any of the other seven?  If the queue is outside your install prefix, aren't either or both of control or users also (potentially) outside the install prefix?

Is the problem that we need to create directories?  If so, can we not just make instchown do that for us?

As presented, this PR is either undersolving a more general packaging problem (separate control of the eight qmail directories), solving a pkgsrc problem in a way that forces other packagers to accommodate it, or has picked the wrong interface (environment variables) to solve the problem as presented.

Given my understanding of the problem (incomplete, as you can see here I have questions), I think you should create an instqueue program.  Some of that work will be frustrating without PR #35, but we made FHS support a 1.08 goal not a 1.07 goal--and FHS seems to be the essential reason this PR is necessary.

mbhangui:
There are advantages of modifying the instpackage to create the queue, namely avoiding duplication of code to write another program to create the queue.
At the same time I totally agree with you that  it is best that we add command line options rather than using env variable. Using an env variable looks like a hack done to fix a problem. I visualize that we may need an instpackage that can install binaries in one directory, man pages in another, queue in another, or just the queue and docs in another (rather than all under /var/qmail).

mbhangui:
If you take a look at ezmlm-0.53, it has an install program which looks at conf-bin and BIN to install binaries and conf-man and MAN to install man pages

The below is from ezmlm makefile
setup: \
it man install conf-bin conf-man
    mkdir -p $(DESTDIR)"`head -n 1 conf-bin`"
    ./install $(DESTDIR)"`head -1 conf-bin`" < BIN
    mkdir -p $(DESTDIR)"`head -n 1 conf-man`"
    ./install $(DESTDIR)"`head -1 conf-man`" < MAN

mbhangui:
At the moment (for 1.07), we may not require these changes. As far as the binary builds are concerned, we can always upload a patch in obs, and make a calal to the patch command in the rpm spec or the debian rules file. Just today @DerDakon  added a patch to address a problem introduced by moving the man pages to the correct FHS location (/usr/share/man). For the next release, I would urge you all to take a look at install.c from ezmlm. It takes input from a text file to install any file in any directory. The only thing missing in that is creating a fifo (which is required in the queue creation - a fifo named trigger), but to add that is just 5 lines change and adding the mkdir_p function from instpackage.c. If we use the install.c from ezmlm, it will eliminate the need for modifying hier.c everytime we add a new file.

schmonz:
> At the moment (for 1.07), we may not require these changes. As far as the binary builds are concerned, we can always upload a patch in obs, and make a call to the patch command in the rpm spec or the debian rules file.

Very true, and that's my backup plan. If we can find a small agreeable change here that makes packaging easier, though, we'll ship a 1.07 that better achieves its goals. (Otherwise we'll have to make do. Either way, once 1.07 is out, we'll need to tackle the FHS problem head-on for 1.08.)

DerDakon:
For the RPMs it looks like simply packaging the queue is fine.

hier.c
```diff
@@ -27,8 +27,11 @@ int mode;
   }
 }
 
-void hier()
+void hier(queue,everythingelse)
+int queue;
+int everythingelse;
```
alanpost: Passing flags as you're doing here can be a so-called 'code smell.'  In this case there is an extra tell: you have a flag for whether or not to install the queue, and a flag for whether or not to not install the queue (With my attendant appreciation in not calling the variable notqmail!).

If one is faced with passing flags to a function it's natural to ask whether the function should refactored instead.  You've added two flags, creating four combinations, one of which makes this function effectively a no-op--such combination reasonably never appearing in the code.  Further, blocks of lines are now if-guarded, which should create a space-only diff in every line--you've taken care here not to do that, but by "squeezing in" an extra level of indent.

Instead of passing flags to hier(), I think you create two functions: hier() and hier_queue().  For hier(1,1) (instcheck, instchown), call both hier() and hier_queue().  For hier(0,1) (instpackage), call hier().  For hier(1,0) (instqueue), call hier_queue().  Pulling hier_queue() out of hier() accomplishes exactly the same thing adding flags does, but in a way that retains the current complexity of this/these subroutines--no extra choices are made at runtime--consistent with these parameters being hard-coded at each calling site.

schmonz: I’d prefer the code in the state you describe, but aimed for an intermediate state that seemed less likely to break patches. I haven’t tested whether that’s true, though.

alanpost: I don't consider the rationale of not breaking patches sufficient justification to put in constant-at-build time flags to a function.  Quite reasonably possible we'll split all of hier() in to the 8 top-level directory-specific functions.  Absolutely impossible this function signature survives the 1.08 series--it is unreasonably overfit for the reasons I describe here.

schmonz: Hmmm. I'm worried we're already slipping down a slope of making common patches non-trivial for users to apply. (For pkgsrc, I've noticed a need to rebase the TLS and SRS patches onto notqmail.) I can't point at data to quantify the barrier that 1.07 will present to users trying to update; I'm just worried about it. But we have ideas for mitigating this, such as the patch branches you've created in the repo. So maybe I need to set my worry aside, watch how it goes for people when they encounter 1.07, and recalibrate accordingly. Let me refactor as you suggest.

schmonz:
On reflection, I'm going to remove my PR here from consideration for 1.07. Reasons:

- The practical need has not been demonstrated outside my own case (pkgsrc), for which I have a workaround or two
- Tackling FHS head-on is slated for 1.08, and that work may make this problem go away
- We are otherwise pretty ready for 1.07

instcheck.c
```diff
@@ -7,6 +7,7 @@
 
 extern void init_uidgid();
 extern void hier();
+extern void hier_queue();
```
DerDakon: Please put that into hier.h, we have one now.

instfiles.c
```diff
@@ -0,0 +1,213 @@
...
+#define FATAL "instfiles: fatal: "
+
+int fdsourcedir = -1;
+
+static void die_nomem()
+{
+  strerr_die2sys(111,FATAL,"out of memory");
+}
+
+static void ddhome(dd,home)
```
DerDakon: C89 API now, please.

josuah:
An `instqueue` program would be useful not only in a post-install script, but also for the admin to manage or repair its mail queue.

Being able to safely repair the mail queue without uninstalling/reinstalling qmail is something I would appreciate to see in qmail.

Maybe even exposing a `qmail-instqueue` with an associated manpage?

I can followup this merge request with anotherone for the renaming and the man page.

schmonz:
@mbhangui, would an `instqueue` program help you with #76? I'd still gladly use it (or something like it) in pkgsrc.

mbhangui:
> @mbhangui, would an `instqueue` program help you with #76? I'd still gladly use it (or something like it) in pkgsrc.

Absolutely. Better that we don't depend on external dependency.

schmonz:
I've rebased this, since we now have

1. More than one package system demonstrating a shared need for something like it
2. No momentum at present for more generally introducing [optional, non-default "FHS-strict" installation](https://github.com/notqmail/notqmail/issues/2#issuecomment-510536117)

I've only been thinking about packagers with this, not sysadmins. @josuah suggests the new queue-initializer is a sensible tool to install and document as `qmail-initqueue` or thereabouts. Do others agree?

hier.c
```diff
@@ -243,3 +227,22 @@ void hier()
   c(auto_qmail,"man/man8","qmail-command.8",auto_uido,auto_gidq,0644);
   c(auto_qmail,"man/cat8","qmail-command.0",auto_uido,auto_gidq,0644);
 }
+
+void hier_queue()
```
DerDakon, line 231: I would put this above ```hier()```, and then just call it from where it was extracted. That way you don't have to modify the files that call ```hier()```, which will behave just as before.

instqueue.c
```diff
@@ -0,0 +1,20 @@
+#include <sys/stat.h>
+#include <stddef.h>
+#include "hier.h"
+#include "open.h"
+#include "strerr.h"
+
+#define FATAL "instqueue: fatal: "
```
DerDakon, line 7: This isn't needed anymore. Put it in front of the message below and call ```strerr_die1sys()```.

DerDakon (changes requested):

instqueue.c
```diff
@@ -0,0 +1,20 @@
...
+
+extern int fdsourcedir;
+
+int main(void)
+{
+  fdsourcedir = open_read(".");
```
DerDakon, line 13: This isn't needed: fdsourcedir is only used in ```c()```, which isn't called from ```hier_queue()```.

Makefile
```diff
@@ -352,8 +352,7 @@ case.h
 
 compile: \
 make-compile warn-auto.sh
-	( cat warn-auto.sh; ./make-compile ) > \
```
DerDakon, line 355: unrelated change

schmonz: It's one of a couple inconsistencies I noticed while preparing `instqueue`. They're in separate commits that seemed small enough and related enough to include here. If we prefer them in a distinct PR or PRs, okay.

schmonz:
When rebasing this recently, I wanted to re-convince myself that no unintentional changes were made (either in the original PR, or in the rebase) in code that moved from `instpackage.c` to `instfiles.c`. Here's a git trick I learned to help with that:

    $ git diff origin/master:instpackage.c instfiles.c

Makefile
DerDakon: Ok, it's intentional, let's just do it. Noone is going to modify that line in a patch anyway.

DerDakon:
Another version is this:

```
$ git blame -C -M instfiles.c
```

leahneukirchen:
When I packaged s/qmail, I rewrote hier.c as a shell script to run at install time. Having a `instqueue` is appreciated as a packager.

leahneukirchen (approved):

DerDakon (approved):

PR #62, opened 2019-08-13
Remove one last reference to SYSDEPS.

schmonz:


alanpost:
This work has already been submitted in PR #33.

PR #67, opened 2019-08-22
remove qmail-upq.

alanpost:
There are qmail queue repair, upgrade, or message removal programs
that modify a queue, and they have more features.


DerDakon:
I guess you forgot "git rm qmail-upq*"?

alanpost:
> I guess you forgot "git rm qmail-upq*"?

Silly enough, yes I did.  Fixed.

PR #71, opened 2019-08-22
Remove the install script.

alanpost:
Use the opportunity of the 2.0 release to remove a backward compatibility thunk, the install script, in favor of directly calling instpackage and instcheck, as our packaging tools already do.


josuah:
If the install script is gone, an install *makefile target* could see the day? #9

Makefile
```diff
@@ -742,10 +742,6 @@ hostname.o: \
 compile hostname.c substdio.h subfd.h substdio.h readwrite.h exit.h
 	./compile hostname.c
 
-install:
-	echo './instpackage && ./instchown' > install
```
josuah, line 746: What about keeping:
```
install:
	./instpackage
```

I do not think that anyone who discovered qmail got used to run `make install` for regenerate this simple shell script, and rather have the `install` target called as a dependency of `make it`.

So we might not break any expectation through your deletion.  Do I open a separate branch for this?

PR #72, opened 2019-08-22
remove HASSHORTSETGROUPS test, use system headers and types instead.

alanpost:
This is the 1.08 work for PR#15, converting the codebase to use uid_t/gid_t and removing the short group test.

@DerDakon has PR #45 which is the same work--only one is needed, and they could we be identical.  I was waiting for the 1.07 release before submitting this PR and now we can sort that out.


prot.c
```diff
@@ -1,21 +1,15 @@
...
-  x[0] = gid; x[1] = 73; /* catch errors */
-  if (setgroups(1,x) == -1) return -1;
-#else
   if (setgroups(1,&gid) == -1) return -1;
-#endif
   return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
 }
 
-int prot_uid(uid) int uid;
+int prot_uid(uid_t uid)
```
DerDakon: I hate to say it, but we are not there yet…

alanpost: > I hate to say it, but we are not there yet…

I had converting K&R function signatures to ANSI (C89) for the 1.08 release.  Did we put that on another release, or you're saying that we should put some 1.08 commits in before we update function signatures?

DerDakon: I would at least separate the "C89 cleanup" and any functional change (and even those adding missing headers) into separate PRs, regardless if they all end up in the same release or not.

prot.c
```diff
@@ -1,21 +1,15 @@
-#include "hasshsgr.h"
+#include <sys/types.h>
+#include <unistd.h>
+#include <grp.h>
 #include "prot.h"
 
-/* XXX: there are more portability problems here waiting to leap out at me */
-
-int prot_gid(gid) int gid;
+int prot_gid(gid) gid_t gid;
```
DerDakon: C89 here and for prot_uid()?

schmonz:
@DerDakon, since we've been dormant a while, can you summarize (for @alanpost, me, and anyone else reading) what's left to address before this looks good to you?

DerDakon:
Look one comment before ;) And then this needs a rebase.

instpackage.c
```diff
@@ -58,8 +59,8 @@ int mode;
 
 void h(home,uid,gid,mode)
```
DerDakon: We could change all of them to C89, too.

DerDakon:
Oops, sorry. I was neither aware that my local branch still pointed to Alans original branch, nor that I have the permissions to override this. Please speak up if we should revert that or if we just want to merge it as it now is. It should address all my previous objections.

schmonz:
> Oops, sorry. I was neither aware that my local branch still pointed to Alans original branch, nor that I have the permissions to override this. Please speak up if we should revert that or if we just want to merge it as it now is. It should address all my previous objections.

I'm having trouble finding time and attention with which to review this. But I'm also realizing I don't feel comfortable being the second approver for this when it's based on @alanpost's work and saved on his branch. (And thank you, @DerDakon, for drawing attention to this unintentional result.)

I would be very comfortable adding a +1 _iff_ Alan's comfortable with (1) the state of the code and (2) the state of the commit history. @alanpost, can you spare some round tuits to review (and meta-review)?

DerDakon:
I pushed this once more to resolve the issues I caused. The first commit is only rebased, and fixes the merge conflicts. These come from functions that have been changed to C89 style meanwhile. The second commit is mine, doing the C89 conversion for every function changed in the first patch that was not C89 style afterwards.

schmonz:
Thank you for clearing this up. I’m comfortable now that Alan’s original work is clearly credited to him. Let’s give it a few more days before merging, just in case he wants to weigh in.

PR #77, opened 2019-08-22
use ANSI/C89 function signatures

alanpost:
This PR is part of the de-K&R series, here adding ANSI (C89) function signatures.

### DIFF
--- a/coe.c
+++ b/coe.c
@@ -1,8 +1,8 @@
+#include <unistd.h>
 #include <fcntl.h>
 #include "coe.h"
 
-int coe(fd)
-int fd;
+int coe(int fd)
 {
   return fcntl(fd,F_SETFD,1);
 }
--- a/qmail-qmqpc.c
+++ b/qmail-qmqpc.c
@@ -135,6 +135,7 @@ char *server;
 
 stralloc servers = {0};
 
+void
 main()
 {
   int i;
--- a/qmail-qmqpc.c
+++ b/qmail-qmqpc.c
@@ -136,7 +136,7 @@ char *server;
 stralloc servers = {0};
 
 void
-main()
+main(void)
 {
   int i;
   int j;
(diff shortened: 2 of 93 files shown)


coe.c
```diff
@@ -1,8 +1,8 @@
+#include <unistd.h>
```
josuah, line 1: - https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html
- http://man.openbsd.org/fcntl
- https://www.freebsd.org/cgi/man.cgi?query=fcntl
- https://netbsd.gw.com/cgi-bin/man-cgi?fcntl++NetBSD-current

None of these state to #include <unistd.h>, but glibc man pages does. Cannot harm to add an extra header...

qmail-qmqpc.c
```diff
@@ -135,7 +135,8 @@ char *server;
 
 stralloc servers = {0};
 
-main()
+void
```
josuah, line 138: I personally use this style, but this has the side effect of shifting the line numbers (does it actually matter for patch?) and is not consistent with the rest of the source, with type and function name on the same line.

quote.h
```diff
@@ -1,8 +1,10 @@
 #ifndef QUOTE_H
 #define QUOTE_H
 
-extern int quote_need();
-extern int quote();
-extern int quote2();
+#include "stralloc.h"
```
josuah, line 4: As a good side effect, we notice more dependencies between source doe files.  Maybe will help to generate the Makefile later...

qmail-qmqpc.c
DerDakon: The line number change will not harm, patch will just output a message about offset.

DerDakon:
Meanwhile we agreed that we will only touch those function signatures that we change for one or the other reason anyway to reduce conflicts with external patches. Since this has it's own amount of collisions I'll close it for now.

PR #75, opened 2019-08-22
lock_*.c, lock.h are now C89.

alanpost:
This PR is part of the de-K&R series, here adding ANSI (C89) function signatures.

PR #79, opened 2019-08-24
include unistd.h in exit.h to declare _exit.

alanpost:
_exit(2) is declared in unistd.h.  Instead of redeclaring it in exit.h, include that system header to get the correct function signature.

As a separate commit, also remove exit.h from forward.c andt maildirwatch.c.  Neither of these programs calls exit directly, and so do not need the exit.h header.

This is the 1.08 portion of the _exit declaration cleanup.  The 1.9 portion of this series is PR #44 .  It depends on PR #80 and so has cherry-picked the single commit from that series.

This change is necessary for -Wall -Werror, to prevent the following kind of error:

qmail-pw2u.c: In function ‘die_read’:
qmail-pw2u.c:34:3: error: incompatible implicit declaration of built-in function ‘_exit’ [-Werror]
   _exit(111);
   ^~~~~

Which cannot be turned off without also turning off -Werror.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1304,7 +1304,7 @@ qmail-pw2u.9 conf-qmail conf-break conf-spawn
 qmail-pw2u.o: \
 compile qmail-pw2u.c substdio.h readwrite.h subfd.h substdio.h \
 sgetopt.h subgetopt.h control.h constmap.h stralloc.h gen_alloc.h \
-fmt.h str.h scan.h open.h error.h getln.h auto_break.h auto_qmail.h \
+fmt.h str.h scan.h open.h error.h getln.h exit.h auto_break.h auto_qmail.h \
 auto_users.h byte.h
 	./compile qmail-pw2u.c
 
@@ -1357,7 +1357,7 @@ qmail-qmtpd.8
 qmail-qmtpd.o: \
 compile qmail-qmtpd.c stralloc.h gen_alloc.h substdio.h qmail.h \
 substdio.h now.h datetime.h str.h fmt.h env.h sig.h rcpthosts.h \
-auto_qmail.h readwrite.h control.h received.h
+auto_qmail.h readwrite.h control.h received.h exit.h
 	./compile qmail-qmtpd.c
 
 qmail-qread: \
--- a/Makefile
+++ b/Makefile
@@ -602,7 +602,7 @@ forward.0: \
 forward.1
 
 forward.o: \
-compile forward.c sig.h readwrite.h exit.h env.h qmail.h substdio.h \
+compile forward.c sig.h readwrite.h env.h qmail.h substdio.h \
 strerr.h substdio.h fmt.h
 	./compile forward.c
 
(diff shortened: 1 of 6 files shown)


DerDakon:
is the readwrite.h change in this PR intentionally?

alanpost:
> is the readwrite.h change in this PR intentionally?

Yes, because it will not build without it.  No, in the sense that PR #80 should just go in before this PR, which will have the side-effect of removing that commit.

I would rather express that problem more cleanly in the GitHub PR interface, but I cannot.

DerDakon:
We could just cherry-pick the 4 commits from #65 that do exactly this, and also do it for fork().

alanpost:
> We could just cherry-pick the 4 commits from #65 that do exactly this, and also do it for fork().

That just moves the problem to PR #65 .  The same problem exists for PR #43 and PR #44, which include commits from PR #80 and PR #79 respectively.

If you would like to pull out the relevant commits to PR #65 as I have done for PR #79 and PR #80 (and PR #81?), feel free.  The problem is that they're not on master and are such low-hanging fruit that a substantial number of the cleanup PRs will depend on them.

I arranged them this way a) at your request and b) to minimize wherever possible the need to cherry-pick.

DerDakon:
My suggestion was more like: do one PR that has the 4 commits from #65, and nothing else. The point of that PR was exactly to decide which parts of it we want now and which later. I guess eventually we want them all.

alanpost:
> do one PR that has the 4 commits from #65, and nothing else.

That does not solve the problem.  PR #43 will still need commit eda3e45  from PR #80, regardless of whether it appears in another PR or as part of a series.  PR #44 will still need commit 5285230 from PR #79, regardless of whether it appears in another PR or as part of a series.  It solves *only* the problem of commit eda3e45 appearing in this PR (#79) and in PR #80 

When I have filed separate PRs (#73, #74, #75), you have asked me to put them together (PR #77).  When I put them together (PRs #43, #44), you have asked me to pull them apart (PR #79 and #44, PR #80 and #43).  As a matter of demonstrated, empirical fact I cannot predict a-priori the size and scope of a commit such that you will not ask for something else.  While it is true that you describe to me your preference after I submit PRs, that is not any apparent guidance I can use to do it better.  PR #65 has this same problem: it does too much to be reviewed at once.  The problem is not just me.

No option available to us satisfies every problem.  Shorter, more focused PRs are categorically easier to review.  There is an apparent *low* limit to the size of a PR, with exception.  The problem is not the PRs, it's that some work is not on master that needs to be on master for individual PRs to be approved.  No arrangement of PRs corrects that issue, only approving them does: hence a larger number of smaller PRs.

DerDakon:
For reference, other have solved this problem in a way causing more trouble IMHO: https://github.com/gentoo/gentoo/blob/master/mail-mta/netqmail/files/1.06-exit.patch

alanpost:
> For reference, other have solved this problem in a way causing more trouble IMHO: https://github.com/gentoo/gentoo/blob/master/mail-mta/netqmail/files/1.06-exit.patch

Look at that.  I put a new commit in this series adding exit.h to these two files.  Note this series also has a couple removals of exit.h.

DerDakon:
Yes, yours does much more than just "get the right declaration".

alanpost:
> Yes, yours does much more than just "get the right declaration".

If you'd like I'll split the header cleanup in to a separate PR--otherwise I'm happy to wait for due diligence and attendant fixes, cleanup, and approval or deferral of that work.

DerDakon:
I would say: do #72 and #108 first, then the last commit is not needed at all.

schmonz:
Bump. @alanpost, how can we help you get this one done?

DerDakon:
Please find my rebased and fixed version branch [pr-include-unistd-exit-h](https://github.com/notqmail/notqmail/tree/pr-include-unistd-exit-h).

PR #80, opened 2019-08-24
include unistd.h in readwrite.h to declare read, write.

alanpost:
read(2) and write(2) are declared in unistd.h.  Instead of redeclaring them in readwrite.h, include that system header to get the correct function signature.

This is the 1.08 portion of the read, write declaration cleanup.  The 1.9 portion of this series is PR #43.

This change is necessary for -Wall -Werror, to prevent the following:

./compile qmail-local.c
In file included from exit.h:4:0,
                 from qmail-local.c:7:
/usr/include/unistd.h:363:16: error: conflicting types for ‘read’
 extern ssize_t read (int __fd, void *__buf, size_t __nbytes) __wur;
                ^~~~
In file included from qmail-local.c:3:0:
readwrite.h:4:12: note: previous declaration of ‘read’ was here
 extern int read();
            ^~~~

Which happens as a side-effect of PR #79, where we include unistd.h everywhere we include exit.h.

### DIFF
--- a/readwrite.h
+++ b/readwrite.h
@@ -1,7 +1,6 @@
 #ifndef READWRITE_H
 #define READWRITE_H
 
-extern int read();
-extern int write();
+#include <unistd.h>
 
 #endif


DerDakon:
For reference: https://github.com/gentoo/gentoo/blob/master/mail-mta/netqmail/files/1.06-readwrite.patch

alanpost:
> For reference: https://github.com/gentoo/gentoo/blob/master/mail-mta/netqmail/files/1.06-readwrite.patch

If folks feel I should, I'll modify the patch to match this one, via removal of the whitespace shown there.  I wasn't aware of this patch prior to submitting this PR, but I've gone ahead and referenced it in the commit since's it's factually prior art.

PR #84, opened 2019-08-25
use ssize_t in substdio

DerDakon:


DerDakon:
I wonder if we should go for C89 function prototypes here if we touch these lines anyway. A collision is a collision, it's not harder to solve with a C89 function prototype than with a K&R one.

alanpost:
> I wonder if we should go for C89 function prototypes here if we touch these lines anyway. A collision is a collision, it's not harder to solve with a C89 function prototype than with a K&R one.

I am broadly wondering the same.  I was going to ask it as "What led you to decide not to also change the type of the various `len` function parameters to size_t?" as I find leaving them and fixing them both persuasive--in some sense it's a question of when.

If you wanted to stage a series of commits showing each step, we can squash them together after review and put them in 1.08, optionally leaving work (not squashing all commits) to 1.9.  I definitely think we're doing ourselves a portability favor by cleaning up this subsystem, now that we're using system-provided function declarations for read and write.

josuah (approved):
It looks like we have two great things here:

factorizing ad-hoc read/write wrappers, and conversion of register native type to ssize_t.

If we are ok with defines, then we reached destination.

If we wat to get rid of the defines later on, maybe using a global variable will do (there is a global variable `qqt`/`qq` already for instance), andthere is always one `struct qmail` per program, so this :

```
#define GEN_QMAILPUT_WRITE(qq) \
ssize_t qmail_put_write(int fd, const void *buf, size_t len) \
{ \
  qmail_put(qq,buf,len); \
  return len; \
}
```

Could become this:

```
ssize_t qmail_put_write(int fd, const void *buf, size_t len)
{
  qmail_put(qqt,buf,len);
  return len;
}
```

I'm doing a patch and testing it after this is merged.

qmail-smtpd.c
```diff
@@ -29,13 +29,7 @@
...
 int timeout = 1200;
 
-ssize_t safewrite(int fd, const void *buf, size_t len)
-{
-  ssize_t r;
-  r = timeoutwrite(timeout,fd,buf,len);
-  if (r == 0 || r == -1) _exit(1);
-  return r;
-}
+GEN_SAFE_TIMEOUTWRITE(safewrite,fd,timeout,_exit(1))
```
schmonz, line 32: Before this commit, `make qmail-smtpd && ./qmail-smtpd` gives `421 unable to read controls (#4.3.0)`. After this commit, it exits 1 silently. I don't understand yet why this change has this effect, but it does.

schmonz: Ah, now I get it. Drop the `fd` argument and have the macro call `timeoutwrite(tout,fd,buf,len)`, not `timeoutwrite(tout,writefd,buf,len)`.

substdio.h
```diff
@@ -1,22 +1,46 @@
...
 typedef struct substdio {
   char *x;
   int p;
   int n;
   int fd;
-  int (*op)();
+  union {
+    ssize_t (*op)(int,void*,size_t);
+    ssize_t (*opw)(int,const void*,size_t);
+  };
```
schmonz, line 15: It appears we could avoid (or defer) the need for `SUBSTDIO_FDBUFW`, and the considerable associated churn, by not introducing these prototypes.

What would we lose (or defer) if we instead replaced `int (*op)()` with `ssize_t (*op)()`?

qmail.c
```diff
@@ -66,11 +66,6 @@ void qmail_put(struct qmail *qq, char *s, size_t len)
   if (!qq->flagerr) if (substdio_put(&qq->ss,s,len) == -1) qq->flagerr = 1;
 }
 
-void qmail_puts(qq,s) struct qmail *qq; char *s;
-{
-  if (!qq->flagerr) if (substdio_puts(&qq->ss,s) == -1) qq->flagerr = 1;
-}
-
```
schmonz, line 73: 👍🏼

getln2.c
```diff
@@ -26,6 +26,8 @@ int sep;
     if (i < n) { substdio_SEEK(ss,*clen = i + 1); *cont = x; return 0; }
     if (!stralloc_readyplus(sa,n)) return -1;
     i = sa->len;
-    sa->len = i + substdio_get(ss,sa->s + i,n);
+    m = substdio_get(ss,sa->s + i,n);
+    if (m != -1)
+      sa->len = i + m;
```
schmonz, line 31: Old bug newly spotted during refactoring? Wondering how you came to discover this, as usual, so I can get better at spotting bugs. But also it looks like nothing in-tree calls `getln2()`. Do patches rely on it, or could we delete?

schmonz (changes requested):
When we merge this PR, I'll be seeing 0 compile warnings on macOS Big Sur. 😃

The abstraction used to be invoked equally for reads and writes, and breaking that symmetry by introducing a write-specific path turns this into rather a large change. I'd like help understanding the tradeoff here.

Prototypes for `read()` and `write()` are indeed distinct, and the abstraction has until now chosen to blur that distinction. It seems the write-specific macro and function become necessary when we decide to make the distinction explicit. What benefits do we expect to gain by doing this? (Inversely, as I noted on `typedef struct substdio`, what would we lose if we chose to continue not specifying our I/O operation's prototype?)

The generator macros for `timeout{read,write}()` need fixing. It's good they don't need an `fd` passed in, because some of the callers were even out of order ;-)

getln2.c
mbhangui: if substdio_feed() has been called before substdio_get(), that condition will not happen. In getln2.c, substdio_feed() is getting called before substdio_get(). Even though that condition will not occur, it is a good practice to to check if substdio_get() returned -1.

qmail-smtpd.c
DerDakon: The macro actually has to be that way: there are several instances where the fd passed to the function is ignored and a local variable is used, e.g. in qmail-remote.c. The bug is that I swapped the arguments to the macro, it must be ```GEN_SAFE_TIMEOUTWRITE(safewrite,timeout,fd,_exit(1))```

getln2.c
DerDakon: This was spotted during the "read may return -2 and actually mean a very large size" refactoring.

schmonz: Guessing we should hit `getln2()` with `deprecated.h`

substdio.h
DerDakon: I was under the impression that this fixes a warning that I have seen somewhere, but it looks like it is actually not the case. Will drop this.

getln2.c
DerDakon: look closer: getln2() is used in getln().

schmonz: Doh! Maybe we should put it (as `static`) in `getln.c`. But unimportant for this PR.

qmail-smtpd.c
schmonz: Yep, I see it now. I may never get my `cpp` license.

substdio.h
schmonz: Woohoo, thank you! This makes a huge difference in the invasiveness of the patch.

substdi.c
```diff
@@ -2,16 +2,10 @@
 #include "byte.h"
 #include "error.h"
 
-static int oneread(op,fd,buf,len)
-register int (*op)();
-register int fd;
-register char *buf;
-register int len;
+static int oneread(ssize_t (*op)(), int fd, char *buf, size_t len)
```
schmonz: `void *buf`?

substdio.c
```diff
@@ -1,11 +1,6 @@
 #include "substdio.h"
 
-void substdio_fdbuf(s,op,fd,buf,len)
-register substdio *s;
-register int (*op)();
-register int fd;
-register char *buf;
-register int len;
+void substdio_fdbuf(substdio *s, ssize_t (*op)(int,void*,size_t), int fd, char *buf, int len)
```
schmonz, line 3: This doesn’t give warnings if `op`’s second arg happens to be `const void *`?

substdio.h
```diff
@@ -1,22 +1,36 @@
...
+#if __STDC_VERSION__ >= 199901L
+#define _QMAIL_HAS_C99
+#endif
+#endif
+
+#ifdef _QMAIL_HAS_C99
+#define SUBSTDIO_FDBUF(o,f,buf,len) { .x = (buf), .p = 0, .n = (len), .fd = (f), .op = (o) }
+#else
+#define SUBSTDIO_FDBUF(o,f,buf,len) { (buf), 0, (len), (f), (o) }
+#endif
```
schmonz, line 26: Do we still need this? The C99 initializer is nicer and safer, of course.

substdo.c
```diff
@@ -3,13 +3,9 @@
 #include "byte.h"
 #include "error.h"
 
-static int allwrite(op,fd,buf,len)
-register int (*op)();
-register int fd;
-register char *buf;
-register unsigned int len;
+static int allwrite(ssize_t (*op)(), int fd, char *buf, size_t len)
```
schmonz, line 6: Fuller `op` signature for symmetry with `oneread`?

schmonz (approved):
Outstanding.

substdio.h
DerDakon: "Need" is a strong word. We can do it. I can also just rip it off, it will make no functional difference.

schmonz: I like the idea, for sure. My vote is it probably makes more sense once we’re ready to mandate C99, as a PR that does this unconditionally for all struct initializers.

substdi.c
DerDakon: Why not ;)

substdio.c
DerDakon: No, because it is (wrongly) just in the C file. I have removed the argument specification in the header, so it has to go away here, too.

DerDakon: In fact, I will leave it as is. See below.

substdo.c
DerDakon: The other way round: I removed the full specification in oneread(), just because we just pass the pointers from substdio in, and that does not have the argument specification.

schmonz (approved):
Looks great. Let’s get this merged!

PR #85, opened 2019-08-25
add missing return types to main()

DerDakon:


alanpost:
A similar treatment (adding explicit void to main) should be given to the following files:

trylsock.c
tryrsolv.c
trysyslog.c

DerDakon:
good catch, no idea why I did not spot them.

trysyslog.c
```diff
@@ -2,8 +2,10 @@
 #include <sys/time.h>
 #include <syslog.h>
 
+int
```
alanpost: why `int` and `return 0` here?  Other try*.c functions (e.g., `tryflock.c`, `trymkffo.c`, `trysalen.c` don't do this, they use void.

DerDakon: but they do _exit(). The return value must come from somewhere.

schmonz: We probably agree that at some point all our `main()`s should return `int`. I'm not understanding why it's better to make that change to these few `try*.c` ones. My thinking is, the problem this PR wants to solve is an internal inconsistency (specifically, that `main()` _almost_ always has _some_ return type). Upgrading to return `int` would make sense if it didn't also add internal consistency (specifically, with the other `try*.c` that already do have a return type). And I don't see any external forces (like the `try*` programs being executed and their exit codes being inspected) to change the balance here.

I appreciate that this is very small and in the right general direction, and that this discussion is far longer than the changes themselves. Still, unless I'm missing something, I think this part of the PR is the seed of another PR.

DerDakon:
The point is: main() is void everywhere, because it does not return and the return code comes from _exit(). I avoid _exit() here to not drag in more headers for no reasons, and just make it ```int main()``` and ```return 0;```.

alanpost:
> The point is: main() is void everywhere, because it does not return and the return code comes from _exit(). I avoid _exit() here to not drag in more headers for no reasons, and just make it `int main()` and `return 0;`.

Nowhere in the codebase does it return from main.  Everywhere it calls _exit.  We should preserve that behavior.

DerDakon:
Btw, Clang now issues these warnings:

```
qmail-qmqpc.c:138:1: warning: return type of 'main' is not 'int'
```

DerDakon:
I'm back to "int main()" now. Fixing one warning with another does not help, and the consistency doesn't give us anything, especially as it's only about try*.c where noone needs to look on any bigger API or structure.

PR #87, opened 2019-08-26
Remove qsmhook, long since replaced by preline.

schmonz:
THOUGHTS says qsmhook is "too bletcherous to be documented. (The official replacement for qsmhook is preline, together with the qmail-command environment variables.)"

I've never used qsmhook, only preline. Guessing nobody has any reason to still need qsmhook.


schmonz:
We announced our [intent to remove in 1.08](https://github.com/notqmail/notqmail/wiki/notqmail-1.08#intent-to-remove), so we're free to do it for 1.09.

PR #89, opened 2019-08-26
Fix leap second handling on systems running TAI

xenotrope:
This patch is intended to fix a minor bug in qmail that uses incorrect
timestamp information on systems using a TAI system clock. Since the TAI
timescale does not permit leap seconds, it is offset from the more
common UTC timescale by an integer number of seconds. As of 2016-12-31,
when another leap second was added, TAI is exactly 37 seconds ahead of
UTC. The 37 seconds results from the initial difference of 10 seconds at
the start of 1972, plus 27 leap seconds in UTC since 1972.

Converting between TAI and UTC is easy if you have an accurate table of
all previous leap seconds, usually placed in "/etc/leapsecs.dat". qmail
does not check for "/etc/leapsecs.dat" and thus does not handle leap
seconds correctly on systems that use TAI without this patch.

Converting qmail's internals to use the libtai library everywhere would
involve rewriting all of its time-handling code. Instead, I chose to
simply scoop out the datetime() function and replace it with a few lines
of libtai's most useful features. This provides support for systems that
use TAI without seriously changing the way qmail understands time.

The file "datetime.c" now contains libtai's ability to convert from
seconds since the UNIX epoch to a caltime struct, using that value to
check "/etc/leapsecs.dat" for the correct TAI - UTC offset. This value
is then given to qmail to digest into its old datetime struct, the
precursor to libtai's caltime.

### DIFF
--- a/caldate.h
+++ b/caldate.h
@@ -1,0 +1,19 @@
+#ifndef CALDATE_H
+#define CALDATE_H
+
+struct caldate {
+  long year;
+  int month;
+  int day;
+} ;
+
+extern unsigned int caldate_fmt(char *, const struct caldate *);
+extern unsigned int caldate_scan(const char *, struct caldate *);
+
+extern void caldate_frommjd(struct caldate *, long, int *, int *);
+extern long caldate_mjd(const struct caldate *);
+extern void caldate_normalize(struct caldate *);
+
+extern void caldate_easter(struct caldate *);
+
+#endif
--- a/caldate_fmjd.c
+++ b/caldate_fmjd.c
@@ -1,0 +1,44 @@
+#include "caldate.h"
+
+void caldate_frommjd(struct caldate *cd, long day, int *pwday, int *pyday)
+{
+  long year;
+  long month;
+  int yday;
+
+  year = day / 146097L;
+  day %= 146097L;
+  day += 678881L;
+  while (day >= 146097L) { day -= 146097L; ++year; }
+
+  /* year * 146097 + day - 678881 is MJD; 0 <= day < 146097 */
+  /* 2000-03-01, MJD 51604, is year 5, day 0 */
+
+  if (pwday) *pwday = (day + 3) % 7;
+
+  year *= 4;
+  if (day == 146096L) { year += 3; day = 36524L; }
+  else { year += day / 36524L; day %= 36524L; }
+  year *= 25;
+  year += day / 1461;
+  day %= 1461;
+  year *= 4;
+
+  yday = (day < 306);
+  if (day == 1460) { year += 3; day = 365; }
+  else { year += day / 365; day %= 365; }
+  yday += day;
+
+  day *= 10;
+  month = (day + 5) / 306;
+  day = (day + 5) % 306;
+  day /= 10;
+  if (month >= 10) { yday -= 306; ++year; month -= 10; }
+  else { yday += 59; month += 2; }
+
+  cd->year = year;
+  cd->month = month + 1;
+  cd->day = day + 1;
+
+  if (pyday) *pyday = yday;
+}
--- a/caltime.h
+++ b/caltime.h
@@ -1,0 +1,20 @@
+#ifndef CALTIME_H
+#define CALTIME_H
+
+#include "caldate.h"
+
+struct caltime {
+  struct caldate date;
+  int hour;
+  int minute;
+  int second;
+  long offset;
+} ;
+
+extern void caltime_tai(const struct caltime *, struct tai *);
+extern void caltime_utc(struct caltime *, const struct tai *, int *, int *);
+
+extern unsigned int caltime_fmt(char *, const struct caltime *);
+extern unsigned int caltime_scan(const char *, struct caltime *);
+
+#endif
--- a/caltime_utc.c
+++ b/caltime_utc.c
@@ -1,0 +1,31 @@
+#include "tai.h"
+#include "leapsecs.h"
+#include "caldate.h"
+#include "caltime.h"
+
+/* XXX: breaks tai encapsulation */
+
+void caltime_utc(struct caltime *ct, const struct tai *t, int *pwday, int *pyday)
+{
+  struct tai t2 = *t;
+  uint64_t u;
+  int leap;
+  long s;
+
+  /* XXX: check for overflow? */
+
+  leap = leapsecs_sub(&t2);
+  u = t2.x;
+
+  u += 58486;
+  s = u % 86400ULL;
+
+  ct->second = (s % 60) + leap; s /= 60;
+  ct->minute = s % 60; s /= 60;
+  ct->hour = s;
+
+  u /= 86400ULL;
+  caldate_frommjd(&ct->date,/*XXX*/(long) (u - 53375995543064ULL),pwday,pyday);
+
+  ct->offset = 0;
+}
--- a/tai.h
+++ b/tai.h
@@ -1,0 +1,28 @@
+/* Public domain. */
+
+#ifndef TAI_H
+#define TAI_H
+
+#include <stdint.h>
+
+struct tai {
+  uint64_t x;
+} ;
+
+#define tai_unix(t,u) ((void) ((t)->x = 4611686018427387914ULL + (uint64_t) (u)))
+
+extern void tai_now(struct tai *);
+
+#define tai_approx(t) ((double) ((t)->x))
+
+extern void tai_add(struct tai *,const struct tai *,const struct tai *);
+extern void tai_sub(struct tai *,const struct tai *,const struct tai *);
+#define tai_less(t,u) ((t)->x < (u)->x)
+
+#define TAI_PACK 8
+extern void tai_pack(char *,const struct tai *);
+extern void tai_unpack(const char *,struct tai *);
+
+extern void tai_uint(struct tai *,unsigned int);
+
+#endif
--- a/tai_unpack.c
+++ b/tai_unpack.c
@@ -1,0 +1,16 @@
+/* Public domain. */
+
+#include "tai.h"
+
+void tai_unpack(const char *s,struct tai *t)
+{
+  uint64_t x;
+  int i;
+
+  x = (unsigned char) s[0];
+  for (i = 1; i < 8; i++) {
+    x <<= 8;
+    x += (unsigned char) s[i];
+  }
+  t->x = x;
+}
--- a/Makefile
+++ b/Makefile
@@ -259,6 +259,14 @@ byte_zero.o: \
 compile byte_zero.c byte.h
 	./compile byte_zero.c
 
+caldate_fmjd.o: \
+compile caldate_fmjd.c caldate.h
+	./compile caldate_fmjd.c
+
+caltime_utc.o: \
+compile caltime_utc.c caltime.h
+	./compile caltime_utc.c
+
 case.a: \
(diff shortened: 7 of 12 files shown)


caldate.h
```diff
@@ -0,0 +1,19 @@
+#ifndef CALDATE_H
+#define CALDATE_H
+
+struct caldate {
+  long year;
```
DerDakon: I'm unsure about the "long" here. We clearly will not have this software in production when an int overflows, but this increases the struct size on 64 bit from 12 to 16 byte. This just fills a cacheline, so probably not really bad, but I'd like to understand why it is "long" and not "int".

caldate.h
```diff
@@ -0,0 +1,19 @@
...
+  int month;
+  int day;
+} ;
+
+extern unsigned int caldate_fmt();
```
DerDakon: F*ck K&R, we should really add proper C89 signatures for new code, especially like this.

caldate_fmjd.c
```diff
@@ -0,0 +1,48 @@
...
+struct caldate *cd;
+long day;
+int *pwday;
+int *pyday;
+{
+  long year;
+  long month;
+  int yday;
+
+  year = day / 146097L;
```
DerDakon: I bet this is correct, but I have no clue what's going on here with all those magic numbers.

caltime_utc.c
```diff
@@ -0,0 +1,35 @@
...
+struct tai *t;
+int *pwday;
+int *pyday;
+{
+  struct tai t2 = *t;
+  uint64 u;
+  int leap;
+  long s;
+
+  /* XXX: check for overfow? */
```
DerDakon: what's "overfow"? ;)

leapsecs_read.c
```diff
@@ -0,0 +1,50 @@
...
+int leapsecs_read()
+{
+  int fd;
+  struct stat st;
+  struct tai *t;
+  int n;
+  int i;
+  struct tai u;
+
+  fd = open("/etc/leapsecs.dat",O_RDONLY | O_NDELAY);
```
DerDakon: I usually do this before:

```c
#ifndef O_CLOEXEC
#define O_CLOEXEC 0
#endif
```

And then pass O_CLOEXEC to all these things just out of paranoia.

leapsecs_read.c
```diff
@@ -0,0 +1,50 @@
...
+  if (fd == -1) {
+    if (errno != ENOENT) return -1;
+    if (leapsecs) free(leapsecs);
```
DerDakon: just free(), it's guaranteed that ```free(NULL)``` does just nothing.

leapsecs_read.c
```diff
@@ -0,0 +1,50 @@
...
+    leapsecs = 0;
+    leapsecs_num = 0;
+    return 0;
+  }
+
+  if (fstat(fd,&st) == -1) { close(fd); return -1; }
+
+  t = (struct tai *) malloc(st.st_size);
```
DerDakon: No, not in any way. If I do just ```echo 0 > /etc/leapsecs.dat" than this will just explode. Better would be to first check that the file size is sensible, and more generally do:

```
t = malloc(sizeof(*t));
```

But I see t neither freed nor returned, so this is probably leaking? And if it is just a local variable, why not just declare it on the stack and scrap the whole heap allocation? This would still need the sanity check on the file size, but removes at least one different error case.

tai_unpack.c
```diff
@@ -0,0 +1,18 @@
+/* Public domain. */
+
+#include "tai.h"
+
+void tai_unpack(const char *s,struct tai *t)
+{
+  uint64 x;
+
+  x = (unsigned char) s[0];
```
DerDakon: This begs for a for loop.

uint64.h1
```diff
@@ -0,0 +1,15 @@
+/* Public domain. */
+
+#ifndef UINT64_H
+#define UINT64_H
+
+/* sysdep: -ulong64 */
```
DerDakon: I would just kill this and ```#include <stdint.h>``` and then use uint64_t, see #30.

caldate_fmjd.c
```diff
@@ -0,0 +1,48 @@
+#include "caldate.h"
+
+void caldate_frommjd(cd,day,pwday,pyday)
```
DerDakon: We agreed on C89 function signatures for things we introduce or touch anyway. Please convert them, and also add them to the header files.

schmonz: @xenotrope, I kinda goaded you into submitting this PR based on your longstanding patch. Are you interested in the possibly tedious work of responding to review by incorporating suggested changes, rebasing and force-pushing your branch, and awaiting further review? Or would you prefer that someone else take it from here? Either way is totally fine. Either way, this is your work and you'll continue to be its git "author" (unless for some reason you'd rather not be, which we can also accommodate).

uint64.h1
DerDakon: We have merged #30 meanwhile, so pleasekill all of that typedefs and just use uint64_t from stdint.h.

leapsecs_read.c
xenotrope: The libtai-0.60 library was designed to be small. It intentionally avoids a lot of the syntactic system libraries that qmail relies upon: open.h, substdio.h/buffer.h, and so on. Perhaps O_CLOEXEC doesn't belong here, but in open_read.c, which would then replace the open() here.

uint64.h1
xenotrope: Done.

leapsecs_read.c
xenotrope: This does not explode. In circumstances where /etc/leapsecs.dat is less than the size of a struct t, n is rounded down to 0 and the *leapsecs pointer is assigned to an empty t the size of st.st_size.

tai_unpack.c
xenotrope: In my testing of a looped unpack function against an unrolled unpack function, the for-loop version was faster than this implementation about 11% of the time. This function is written the way it is for performance reasons. (I imagine rewriting it as a loop would cause it to get unrolled as a compiler optimization anyway; if so maybe the compiler could make it faster.)

uint64.h1
DerDakon: not the second part ;)

xenotrope: This commit was pushed to be in alignment with PR #30 and the uint32.h header file which typedefs the uint32 datatype to uint32_t. Replacing uint64 entirely in favor of uint64_t can also be done once I walk through all of the tai_*.[ch] files and fix the includes.

DerDakon: We have not done this in #30 as it would cause collisions with other patches as many files are touched. Here all users of that type are your new code anyway so there is nothing this can collide with.

tai_unpack.c
DerDakon: Compilers are clever beasts these days, they could even use SIMD instructions or similar things if they think it's useful.

leapsecs_read.c
DerDakon: One needs the file descriptor in a child process e.g. when passing the mail body to qmail-remote, so always using O_CLOEXEC is probably wrong.

leapsecs_read.c
DerDakon: Yes, you are right, but I still find it dangerous, and also unnecessary.

 * if it's smaller than struct tai one can immediately leave the function and not do any allocation at all as it would be ignored anyway
 * if it's bigger than struct tai it is completely read into memory, even if it is _big_, so it could DoS you if you drop a wrong file there by accident. I would mmap() it if one really needs the whole content.

leapsecs_read.c
```diff
@@ -0,0 +1,49 @@
...
+int leapsecs_read()
+{
+  int fd;
+  struct stat st;
+  struct tai *t;
+  int n;
+  int i;
+  struct tai u;
+
+  fd = open("/etc/leapsecs.dat",O_RDONLY | O_NDELAY);
```
josuah, line 20: Adding this new file as a resource, I would be useful to have a documentation, maybe on the wiki to signal this change of behavior.  Containers using qmail would, for instance, lack the file, or the system may simply not have the tai file installed.

Does this mean we need to ship the tools to maintain this `/etc/leapsecs.dat` or do we let libtai do that work?

Hopefully I can come-up with something simple that converts this https://hpiers.obspm.fr/iers/bul/bulc/ntp/leap-seconds.list to the libtai format.

schmonz:
Now that 1.08 is out, I'd like to see this (along with #90) rebased and merged sometime soon. I've assigned it the 1.09 milestone. @xenotrope, might you have attention for this in the next week or so?

leapsecs_read.c
```diff
@@ -0,0 +1,48 @@
+#include <sys/stat.h>
+#include <errno.h>
+extern int errno;
```
DerDakon, line 3: This is missing <unistd.h> for close() and friends.

caldate_fmjd.c
DerDakon: just writing ```(365 * 400)``` would be a start, as every existing compiler will just do the multiplication at compile time anyway, but a comment what this was all about (if someone knows) is even better.

leapsecs_read.c
DerDakon: And also delete the "extern int errno;" now that it includes <errno.h>

leapsecs_read.c
```diff
@@ -0,0 +1,48 @@
...
+  struct tai *t;
+  int n;
+  int i;
+  struct tai u;
+
+  fd = open_read("/etc/leapsecs.dat");
+  if (fd == -1) {
+    if (errno != ENOENT) return -1;
+    if (leapsecs) free(leapsecs);
+    leapsecs = 0;
```
DerDakon, line 23: ```NULL``` please, this is a pointer.

leapsecs_read.c
DerDakon: This also does no need a cast if the malloc() would be kept, as void* can be assigned to any other pointer type in C without cast. May come from a time where it was alloc(), which returned char*.

leapsecs_read.c
DerDakon: This also misses an ```#include <stdlib.h>``` as it uses free().

tai.h
```diff
@@ -0,0 +1,28 @@
+/* Public domain. */
```
DerDakon, line 1: Can we replace these with the equivalent SPDX line? Would help any tooling that may eventually look at the license of the files.

DerDakon (changes requested):

caltime.h
```diff
@@ -0,0 +1,20 @@
...
+#define CALTIME_H
+
+#include "caldate.h"
+
+struct caltime {
+  struct caldate date;
+  int hour;
+  int minute;
+  int second;
+  long offset;
```
DerDakon, line 11: This is used exactly once, in caltime_utc.c where it is initialized with 0, which makes me think we can just remove it.

tai_unpack.c
```diff
@@ -0,0 +1,16 @@
+/* Public domain. */
+
+#include "tai.h"
+
+void tai_unpack(const char *s,struct tai *t)
```
DerDakon, line 5: I think the interface of this function is unfortunate. When called it is called with a struct tai, so the pointer needs to be casted. But internally this casts again. So the interface should either take a unsigned char* directly, or just take void* and cast it to an unsigned char* internally that can be used without continuous casting. I'd slightly prefer the latter.

PR #90, opened 2019-08-26
Improve Maildir file uniqueness in qmail-local.c

xenotrope:
Some operating systems quickly recycle PIDs, which can lead to
collisions between Maildir-style filenames, which must be unique and
non-repeatable within one second. This patch is just a means of updating
qmail-local to use the format of the revised Maildir protocol.

It uses four unique identifiers:
  * inode number of the file written to Maildir/tmp
  * device number of the file written to Maildir/tmp
  * time in microseconds
  * the PID of the writing process

A Maildir-style filename would look like the following:

In Maildir/tmp:
  time.MmicrosecondsPpid.host
In Maildir/new:
  time.IinodeVdeviceMmicrosecondsPpid.host

Additionally, this patch further comforms to the revised Maildir
protocol by looking through the hostname for instances of '/' and ':',
replacing them with "\057" and "\072", respectively, when writing it to
disk. Special thanks go to Matthias Andree for design & sanity-checking.

### DIFF
--- a/fmt_xlong.c
+++ b/fmt_xlong.c
@@ -1,0 +1,17 @@
+#include "fmt.h"
+
+/* writes ulong u in hex to char *s, does not NULL-terminate */
+unsigned int fmt_xlong(char *s, unsigned long u)
+{
+ unsigned int len;
+ unsigned long q;
+ unsigned long c;
+ len = 1; q = u;
+ while (q > 15) { ++len; q /= 16; }
+ if (s)
+  {
+   s += len;
+   do { c = u & 15; *--s = (c > 9 ? 'a' - 10 : '0') + c; u /= 16; } while(u);
+  }
+ return len;
+}
--- a/qmail-local.c
+++ b/qmail-local.c
@@ -1,4 +1,5 @@
 #include <sys/types.h>
+#include <sys/time.h>
 #include <sys/stat.h>
 #include <stdlib.h>
 #include "readwrite.h"
@@ -65,6 +66,7 @@ stralloc ueo = {0};
 stralloc cmds = {0};
 stralloc messline = {0};
 stralloc foo = {0};
+stralloc hostname = {0};
 
 char buf[1024];
 char outbuf[1024];
@@ -80,7 +82,7 @@ void maildir_child(dir)
 char *dir;
 {
  unsigned long pid;
- unsigned long time;
+ struct timeval time;
  char myhost[64];
  char *s;
  int loop;
@@ -88,27 +90,47 @@ char *dir;
  int fd;
  substdio ss;
  substdio ssout;
+ stralloc hostsa = {0};
+ size_t hostnamemaxlen = 256;
 
  sig_alarmcatch(sigalrm);
  if (chdir(dir) == -1) { if (error_temp(errno)) _exit(1); _exit(2); }
- pid = getpid();
  myhost[0] = 0;
- gethostname(myhost,sizeof(myhost));
+ pid = getpid();
+ if (!stralloc_ready(&hostsa,hostnamemaxlen)) temp_nomem();
+ if (!gethostname(hostsa.s,hostnamemaxlen)) error_temp(errno);
+ hostsa.s[hostnamemaxlen-1] = 0;
+
+ s = hostsa.s;
+ for (loop = 0; loop < str_len(hostsa.s); ++loop)
+  {
+   if (hostsa.s[loop] == '/')
+    {
+     if (!stralloc_cats(&hostname,"\\057")) temp_nomem();
+     continue;
+    }
+   if (hostsa.s[loop] == ':')
+    {
+     if (!stralloc_cats(&hostname,"\\072")) temp_nomem();
+     continue;
+    }
+   if (!stralloc_append(&hostname,s+loop)) temp_nomem();
+  }
+
  for (loop = 0;;++loop)
   {
-   time = now();
+   gettimeofday(&time, NULL);
    s = fntmptph;
    s += fmt_str(s,"tmp/");
-   s += fmt_ulong(s,time); *s++ = '.';
-   s += fmt_ulong(s,pid); *s++ = '.';
-   s += fmt_strn(s,myhost,sizeof(myhost)); *s++ = 0;
+   s += fmt_ulong(s,time.tv_sec); *s++ = '.';
+   *s++ = 'M'; s += fmt_ulong(s,time.tv_usec);
+   *s++ = 'P'; s += fmt_ulong(s,pid); *s++ = '.';
+   s += fmt_strn(s,hostname.s,hostname.len); *s++ = 0;
    if (stat(fntmptph,&st) == -1) if (errno == error_noent) break;
    /* really should never get to this point */
    if (loop == 2) _exit(1);
    sleep(2);
   }
- str_copy(fnnewtph,fntmptph);
- byte_copy(fnnewtph,3,"new");
 
  alarm(86400);
  fd = open_excl(fntmptph);
@@ -126,8 +148,23 @@ char *dir;
   }
 
  if (substdio_flush(&ssout) == -1) goto fail;
+ if (fstat(fd, &st) == -1) goto fail;
  if (fsync(fd) == -1) goto fail;
  if (close(fd) == -1) goto fail; /* NFS dorks */
+
+ s = fnnewtph;
+ s += fmt_str(s,"new/");
+ s += fmt_ulong(s,time.tv_sec); *s++ = '.';
+
+ /* in hexadecimal */
+ *s++ = 'I'; s += fmt_xlong(s,st.st_ino);
+ *s++ = 'V'; s += fmt_xlong(s,st.st_dev);
+
+ /* in decimal */
+ *s++ = 'M'; s += fmt_ulong(s,time.tv_usec);
+ *s++ = 'P'; s += fmt_ulong(s,pid); *s++ = '.';
+
+ s += fmt_strn(s,hostname.s,hostname.len); *s++ = 0;
 
  if (link(fntmptph,fnnewtph) == -1) goto fail;
    /* if it was error_exist, almost certainly successful; i hate NFS */
--- a/Makefile
+++ b/Makefile
@@ -568,6 +568,10 @@ fmt_ulong.o: \
 compile fmt_ulong.c fmt.h
 	./compile fmt_ulong.c
 
+fmt_xlong.o: \
+compile fmt_xlong.c fmt.h
+	./compile fmt_xlong.c
+
 fmtqfn.o: \
 compile fmtqfn.c fmtqfn.h fmt.h auto_split.h
 	./compile fmtqfn.c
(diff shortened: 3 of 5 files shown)


mbhangui:
fmt_xlong() should be moved to fmt_xlong.c.
we should add the fmt_xlong() proto to fmt.h

qmail-local.c
```diff
@@ -92,21 +108,37 @@ char *dir;
  pid = getpid();
  host[0] = 0;
  gethostname(host,sizeof(host));
+
```
DerDakon: one could do an initial allocation for hostname here, so it has enough room to hold "host" first, which should be enough for all sane cases anyway. Or maybe just do a scan through host if it has any of the special chars and just copy over the whole string otherwise (i.e. always).

qmail-local.c
```diff
@@ -41,6 +42,20 @@ void temp_slowlock()
 void temp_qmail(fn) char *fn;
 { strerr_die5x(111,"Unable to open ",fn,": ",error_str(errno),". (#4.3.0)"); }
 
+/* writes ulong u in hex to char *s, does not NULL-terminate */
+unsigned int fmt_xlong(s,u) char *s; unsigned long u;
```
DerDakon: C89 signatures, please.

schmonz: @xenotrope, as with #89, please let us know your preference to shepherd this through to approval and merge to master, or to back slowly away and have someone else handle that. Either way is totally fine.

xenotrope: fmt_xlong() has been moved into its own file and compiled as part of the fs.a library along with the other fmt_* functions.

fmt_xlong.c
```diff
@@ -0,0 +1,13 @@
+/* writes ulong u in hex to char *s, does not NULL-terminate */
```
DerDakon: Please include fmt.h here so we are sure that the signatures match.

schmonz:
I'd like to see this rebased and merged sometime soon after 1.08 is out, so I've assigned it the 1.09 milestone. @xenotrope, might you have attention for this sometime in the next week-ish?

fmt_xlong.c
```diff
@@ -0,0 +1,15 @@
+#include "fmt.h"
+
+/* writes ulong u in hex to char *s, does not NULL-terminate */
+unsigned int fmt_xlong(char *s, unsigned long u)
+{
+ unsigned int len; unsigned long q; unsigned long c;
```
DerDakon, line 6: As I was forced to unwrap the gen_allocdefs code recently: please put only one instruction per line, and only one type of variable.

qmail-local.c
leahneukirchen: POSIX gethostname does not null-terminate the buffer if it was too long, so I recommend:
```
        host[sizeof host - 1] = 0;
```

xenotrope: The best solution here is for users to adopt OpenBSD's gethostname(3): "The returned name is always NUL terminated." I've updated this to use a stralloc and then store the hostname in the sa.s array and write '\0' in the last byte. For hostnames with 256 or more characters, this will truncate it to the first 255.

qmail-local.c
```diff
@@ -92,21 +94,37 @@ char *dir;
...
+     if (!stralloc_cats(&hostname,"\\072")) temp_nomem();
+     continue;
+    }
+   if (!stralloc_append(&hostname,s+loop)) temp_nomem();
+  }
+
  for (loop = 0;;++loop)
   {
-   time = now();
+   gettimeofday(&time, 0);
```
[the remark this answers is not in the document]
xenotrope: Done.

qmail-local.c
xenotrope: fmt_xlong.c has has its len, q, and c variables broken into one-per-line to match the changes being made to genalloc.

fmt_xlong.c
xenotrope: Done.

DerDakon:
You now add ```fmt_xlong()``` back to ```fmt.h``` after we removed it in #126. Please add it with complete function signature.

qmail-local.c
```diff
@@ -79,35 +95,54 @@ void maildir_child(dir)
...
  for (loop = 0;;++loop)
   {
-   time = now();
+   gettimeofday(&time, NULL);
    s = fntmptph;
    s += fmt_str(s,"tmp/");
-   s += fmt_ulong(s,time); *s++ = '.';
-   s += fmt_ulong(s,pid); *s++ = '.';
-   s += fmt_strn(s,myhost,sizeof(myhost)); *s++ = 0;
+   s += fmt_ulong(s,time.tv_sec); *s++ = '.';
```
DerDakon, line 137: This adds both '.' and 'M', I don't thinkg that is intentional.

qmail-local.c
DerDakon: Conversion to C89 still missing.

qmail-local.c
```diff
@@ -42,6 +43,20 @@ void temp_slowlock()
 void temp_qmail(fn) char *fn;
 { strerr_die5x(111,"Unable to open ",fn,": ",error_str(errno),". (#4.3.0)"); }
 
+/* writes ulong u in hex to char *s, does not NULL-terminate */
+unsigned int fmt_xlong(s,u) char *s; unsigned long u;
+{
+ unsigned int len; unsigned long q; unsigned long c;
```
DerDakon, line 49: Please unwrap them also, which also gives the opportunity to initialize them immediately and get rid of the next line entirely.

qmail-local.c
DerDakon: This still misses to preallocate hostname as shown above.

One could also think of not preallocating and simply not touching hostname at all until we actually found a position where it needs to be recoded, and only then allocate it and copy the whole prefix at once. This would however make the code slightly more complicated.

qmail-local.c
```diff
@@ -79,35 +95,54 @@ void maildir_child(dir)
 char *dir;
 {
  unsigned long pid;
- unsigned long time;
- char myhost[64];
+ struct timeval time;
+ char host[64];
```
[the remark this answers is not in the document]
schmonz: P.S. I recognize we’ve made this PR more labor-intensive for you by changing a bunch of stuff out from under it. Sorry about this! And please keep going — I think we’re pretty close to making maintenance of this code _everyone’s_ job.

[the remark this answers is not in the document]
xenotrope: Hopefully maintaining this code doesn't become necessary for anyone, at least not in a long-term capacity. While I deeply appreciate your apology, (really, thank you!) it's not necessary. A number of these new changes seem[1] to be because the "git rebase master" undid some commits I've made in an earlier version of this branch to address previous coding concerns. I blame myself. And also Junio Hamano. Any chance somebody has the temerity to make "darcshub.com" or "Codevillehub.com"?

[1] I'm not entirely sure, because everything seemed like it was OK before I pushed, but... my mother said that once and she was gravely mistaken.

qmail-local.c
```diff
@@ -42,6 +43,20 @@ void temp_slowlock()
 void temp_qmail(fn) char *fn;
 { strerr_die5x(111,"Unable to open ",fn,": ",error_str(errno),". (#4.3.0)"); }
 
+/* writes ulong u in hex to char *s, does not NULL-terminate */
```
DerDakon, line 46: This function now lives in it's own file, so drop it here.

qmail-local.c
xenotrope: The old Maildir format is "time.pid.hostname:flags". By replacing "pid" with more stuff, including a sub-second time measurement, this patch changes it to "time.MμsecPpid.host:flags". So time.tv_sec and time.tv_usec are both needed, but one comes before the first dot and the other comes after it.

qmail-local.c
xenotrope: fmt_xlong() was removed from qmail-local.c a while back and this regression will be fixed in a new push.

fmt_xlong.c
```diff
@@ -0,0 +1,17 @@
...
+
+/* writes ulong u in hex to char *s, does not NULL-terminate */
+unsigned int fmt_xlong(char *s, unsigned long u)
+{
+ unsigned int len;
+ unsigned long q;
+ unsigned long c;
+ len = 1; q = u;
+ while (q > 15) { ++len; q /= 16; }
+ if (s)
```
DerDakon, line 11: Please reformat this block: put the { in the line of the if() and unwrap the do-while loop.

qmail-local.c
```diff
@@ -80,35 +82,55 @@ void maildir_child(dir)
...
- pid = getpid();
  myhost[0] = 0;
- gethostname(myhost,sizeof(myhost));
+ pid = getpid();
+ if (!stralloc_ready(&hostsa,hostnamemaxlen)) temp_nomem();
+ if (!gethostname(hostsa.s,hostnamemaxlen)) error_temp(errno);
+ hostsa.s[hostnamemaxlen-1] = 0;
+
+ s = hostsa.s;
+ for (loop = 0; loop < str_len(hostsa.s); ++loop)
```
DerDakon, line 105: please use ```strlen()``` instead of ```str_len()``` in new code.

qmail-local.c
DerDakon: Please put all of them in different lines to improve readability. If you want to group the sections than insert newlines.

DerDakon (changes requested):

qmail-local.c
```diff
@@ -80,35 +82,55 @@ void maildir_child(dir)
...
+    {
+     if (!stralloc_cats(&hostname,"\\057")) temp_nomem();
+     continue;
+    }
+   if (hostsa.s[loop] == ':')
+    {
+     if (!stralloc_cats(&hostname,"\\072")) temp_nomem();
+     continue;
+    }
+   if (!stralloc_append(&hostname,s+loop)) temp_nomem();
```
josuah, line 117: Just making sure: there is no program in qmail that tries to decode these escape codes?
It is just in case any of those are encountered?
Why not hard fail instead? It is not expected to have these characters in a host name in the first place, so it might not bother anyone if it stops in these cases.

>       <name>  ::= <let>[*[<let-or-digit-or-hyphen>]<let-or-digit>]
> https://datatracker.ietf.org/doc/html/rfc952

Although, it *may* happen: `hostname /` works just fine here.

qmail-local.c
josuah: Well-behaving mail software would ignore all the pseudo-random part of the name, so I suppose adding letters in addition to `.` and `0-9` is fine.
Also, I think I remember there are already software around using letters in there.

josuah (approved):
Nothing to add to what @DerDakon and @leahneukirchen said.

qmail-local.c
leahneukirchen: The encoding of / and : is specified in https://cr.yp.to/proto/maildir.html, there's no need to decode it.

schmonz:
@xenotrope, would you be comfortable with me handling the rebasing, etc. and getting this PR merged? I can make sure to keep the change attributed to you (as the patch itself is). We want this one in 1.09 and I'd be happy to take it from here, ideally with your permission.

aaeegg:
The wall-clock time from gettimeofday can sometimes be non-monotonic. On systems that have a monotonic uptime count readable by clock_gettime(CLOCK_MONOTONIC), how about putting that in the delivery identifier in addition to the gettimeofday time?

I propose expressing this as O*nm*, where _n_ is (in hexadecimal) the tv_sec from clock_gettime(CLOCK_MONOTONIC) and _m_ is (in hexadecimal, zero-padded on the left to 8 digits) the tv_nsec.

PR #92, opened 2019-08-27
qmail-pop3d: exit 1, pronto, if running as root.

schmonz:
At least checkpassword-pam and DJB's original checkpassword are willing
to authenticate the root user and run a child program with UID 0. For
POP3 service -- the canonical checkpassword use case! -- this produces
runtime behavior that is inconsistent with...

- qmail's design, because qmail-local will never run as root to populate
  a Maildir that qmail-pop3d could serve, and

- qmail's security focus, because it lets abusers dictionary-attack the
  root password over the network.

Instead, in qmail-pop3d, if getuid() returns 0, we write to the logs and
exit 1. If someone manages to guess the root password, they won't know
they did: it looks exactly like any other failed checkpassword login.
And the sysadmin could potentially be alerted.

### DIFF
--- a/qmail-pop3d.c
+++ b/qmail-pop3d.c
@@ -1,5 +1,6 @@
 #include <sys/types.h>
 #include <sys/stat.h>
+#include <unistd.h>
 #include "commands.h"
 #include "sig.h"
 #include "getln.h"
@@ -35,6 +36,9 @@ int safewrite(fd,buf,len) int fd; char *buf; int len;
   return r;
 }
 
+char sserrbuf[128];
+substdio sserr = SUBSTDIO_FDBUF(safewrite,2,sserrbuf,sizeof sserrbuf);
+
 char ssoutbuf[1024];
 substdio ssout = SUBSTDIO_FDBUF(safewrite,1,ssoutbuf,sizeof ssoutbuf);
 
@@ -63,6 +67,10 @@ void err(s) char *s;
 
 void die_nomem() { err("out of memory"); die(); }
 void die_nomaildir() { err("this user has no $HOME/Maildir"); die(); }
+void die_root() {
+  substdio_putsflush(&sserr,"qmail-pop3d invoked as uid 0, terminating\n");
+  _exit(1);
+}
 void die_scan() { err("unable to scan $HOME/Maildir"); die(); }
 
 void err_syntax() { err("syntax error"); }
@@ -294,6 +302,7 @@ char **argv;
   sig_alarmcatch(die);
   sig_pipeignore();
  
+  if (!getuid()) die_root();
   if (!argv[1]) die_nomaildir();
   if (chdir(argv[1]) == -1) die_nomaildir();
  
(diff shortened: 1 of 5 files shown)


qmail-pop3d.c
```diff
@@ -294,6 +302,7 @@ char **argv;
   sig_alarmcatch(die);
   sig_pipeignore();
  
+  if (!getuid()) die_root();
```
alanpost: getuid is defined in unistd.h, whouch should be included right below the sys/ headers.

qmail-popup.c
```diff
@@ -88,7 +88,6 @@ char *pass;
   int wstat;
   int pi[2];
  
-  if (fd_copy(2,1) == -1) die_pipe();
```
alanpost: I think I understand why you did this here, but will you explain it so I'm sure?  If it's necessary the qmail-popup.8 manpage needs to be corrected, as it explains how these file descriptors are used.

schmonz: Totally missed that in the manpage. I did this proximately because pop3d couldn't write a log entry by writing to stderr, and ultimately because I've never understood why the interface was defined that way. [qmail.org's Tips & Advice](http://qmailorg.schmonz.com/top.html#tips) suggests this nutty workaround:

> Qmail-popup redirects stderr to stdout, thus making it impossible to write a wrapper around qmail-pop3d which writes to the logfile by writing to stderr. Being a little cleverer with the shell, you can also redirect FD 7 onto stdout like this:
> 
> `/var/qmail/bin/qmail-pop3d-wrapper.sh /var/qmail/bin/qmail-pop3d Maildir 2>&1 7>&1`
> 
> Once you've done that, qmail-pop3d-wrapper.sh can log to FD 7, like this:
> 
> `#!/bin/sh`
> `echo "qmail-pop3d: user $USER logged in from $TCPREMOTEIP:$TCPREMOTEPORT" >&7`
> `$@`

So I changed the popup interface to make it possible to write a log entry the same way everything else in qmail does. If we agree that on that change, I'll update the popup manpage to match.

alanpost: > I changed the popup interface to make it possible to write a log entry the same way everything else in qmail does. If we agree that on that change, I'll update the popup manpage to match.

It may not actually, literally be "everything else," but your point is well taken--the exceptions if they exist probably support your point rather than contradict it.

Would you make this work a separate commit in this PR?  I'll need to do a run through of the code for this behavior elsewhere, for due diligence, which may scare up other material to record, change, or accommodate, but on it's own merit this change is reasonable and superior to the workaround.

qmail-pop3d.c
```diff
@@ -63,6 +67,11 @@ void err(s) char *s;
 
 void die_nomem() { err("out of memory"); die(); }
 void die_nomaildir() { err("this user has no $HOME/Maildir"); die(); }
+void die_root() {
+  substdio_puts(&sserr,"qmail-pop3d invoked as uid 0, terminating\n");
```
DerDakon, line 71: You could use ```substdio_putflush()```, which can do this with a single call. Since it's basically unused you can then reduce the size of sserrbuf, or even completely remove it.

PR #93, opened 2019-08-27
remove inefficient maildirwatch

DerDakon:
This made sense at a time where maildir was a new concept, but most relevant mailers natively support maildir these days. There are also more efficient syscalls to watch directories meanwhile, and people usually also do not have local mail stores anyway.


schmonz:
> we should publish our intent in advance

Second (maybe third) thoughts: meh. Maybe that'll be important when we have more users. For now, if we want to delete, let's delete.

CHANGES
```diff
@@ -1,3 +1,4 @@
+20190827 remove maildirwatch, you can do this more efficiently using *notify() syscalls
```
schmonz: maybe just "remove inefficient maildirwatch"?

DerDakon: Yes, the rest could be in the commit message.

schmonz:
> > we should publish our intent in advance
> 
> Second (maybe third) thoughts: meh. Maybe that'll be important when we have more users. For now, if we want to delete, let's delete.

Fourth thoughts, after IRC discussion with @alanpost: Given our claim to be conservative and careful, if we remove a program that someone was relying on without giving them a chance to change our minds, they will consider our claim false. It's expensive to have users perceiving us as rash; it's cheap to wait a little longer before deleting (or treating as extensions, which might feel pretty similar).

DerDakon:
fine with me.

PR #102, opened 2019-09-10
Add a test target and one unit test, using Check.

schmonz:
Had a little time on the plane. Here's an extremely minimal example of a unit test. Not looking for approval yet, just feedback.

You'll need [Check](https://libcheck.github.io/check/doc/check_html/index.html#Top) installed, and depending where it goes you may need to adjust `conf-cc` and `conf-ld` to find it.

Then run `make test` and behold, at least one use of stralloc is shown to work!

I picked stralloc because we all understand it, it's heavily used, it's relatively easy to test, and who knows, maybe with sufficient stralloc tests we'll discover something interesting we'll be glad to know about.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ NROFF=nroff
 
 default: it
 
-.PHONY: check clean default it man
+.PHONY: check clean default it man test
 
 %.0:
 	$(NROFF) -man $^ > $@
@@ -352,6 +352,7 @@ exit.h auto_spawn.h
 clean: \
 TARGETS
 	rm -f `cat TARGETS`
+	$(MAKE) -C tests clean
 
 coe.o: \
 compile coe.c coe.h
@@ -2002,6 +2003,9 @@ byte.h datetime.h readwrite.h
 tcpto_clean.o: \
 compile tcpto_clean.c tcpto.h open.h substdio.h readwrite.h
 	./compile tcpto_clean.c
+
+test: it
+	@$(MAKE) -C tests test
 
 timeoutconn.o: \
 compile timeoutconn.c ndelay.h select.h error.h readwrite.h ip.h \
--- a/tests/unittest_stralloc.c
+++ b/tests/unittest_stralloc.c
@@ -1,0 +1,51 @@
+#include <check.h>
+
+#include "alloc.h"
+#include "stralloc.h"
+
+START_TEST(test_stralloc_thingy)
+{
+  stralloc thingy = {0};
+  const char *input = "thingy";
+
+  stralloc_copys(&thingy,input);
+  stralloc_0(&thingy);
+
+  ck_assert_str_eq(input, thingy.s);
+
+  alloc_free(thingy.s);
+}
+END_TEST
+
+TCase
+*stralloc_something(void)
+{
+  TCase *tc = tcase_create("basic operations");
+
+  tcase_add_test(tc, test_stralloc_thingy);
+
+  return tc;
+}
+
+Suite
+*stralloc_suite(void)
+{
+  Suite *s = suite_create("notqmail stralloc");
+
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -2074,6 +2078,15 @@ uid.o: \
 compile uid.c uidgid.h subfd.h substdio.h exit.h
 	./compile uid.c
 
+unittest_stralloc: \
```
DerDakon: Please put these into a subdirectory: it will cause much less jitter in the main Makefile, causing less collisions with other patches and PRs.

unittest_stralloc.c
```diff
@@ -0,0 +1,48 @@
+#include <check.h>
+
+#include "stralloc.h"
+
+START_TEST(test_stralloc_thingy)
+{
+  stralloc thingy = {0};
+  char *input = "thingy";
```
DerDakon: const? Depends on if it causes a warning when being const, but should usually be a good idea.

Makefile
schmonz: I'm in favor of avoiding those collisions, but I'm not sure how best to handle the dependencies in a separate dir and/or Makefile. If there's a way you might like, would you append a commit and push to this branch so I can see what you mean?

schmonz:
I believe this branch is now merge-ready. It's one small test for a `stralloc`, one giant leap for giving us a clear way to keep adding tests.

Before building and running the tests, install [Check](https://libcheck.github.io/check/) (and [pkg-config](https://pkg-config.freedesktop.org) or [pkgconf](https://github.com/pkgconf/pkgconf) to find Check's includes and libs). These are dependencies for tests and only for tests.

To build and run the tests from the top-level `Makefile`, simply `make test`. Verified to work with recent BSD and GNU `make`.

To add more `stralloc` tests, copy the `test_stralloc_thingy` example in `tests/unittest_stralloc.c`. To add other tests, copy the `unittest_stralloc` example in `tests/Makefile`.

Travis and Cirrus now run `make test`, too.

In our too-brief meetup on Friday, @DerDakon suggested running tests via `cmake` to reduce build complexity, generate test reports, and possibly help in other ways. I'm open to it, but I'd love to see us first write a bunch of tests and learn what's annoying about that. :-)

The sooner we merge this, the sooner we can start adding tests with most every commit. The sooner we have existing code covered by test, the sooner we can make bolder changes. With good test coverage we can also easily get good `valgrind` coverage. So I'd like to see this go in ASAP, with the understanding that the way we write, run, and use the results of tests will necessarily need to be refined over time.

tests/unittest_stralloc.c
```diff
@@ -0,0 +1,48 @@
...
+START_TEST(test_stralloc_thingy)
+{
+  stralloc thingy = {0};
+  const char *input = "thingy";
+
+  stralloc_copys(&thingy,input);
+  stralloc_0(&thingy);
+
+  ck_assert_str_eq(input, thingy.s);
+}
```
DerDakon: This leaks the contents of thingy.

schmonz: You're right. Initially I wasn't sure this was worth doing, because in the entire source tree nearly every `sa.s` relies on the process to end, rather than explicitly calling `alloc_free()` when the stralloc goes out of scope. But as @DerDakon reminded me in IRC, we know we'll want to run tests under `valgrind`.

While I'm mentioning @DerDakon, I've neglected to mention that in this revised PR submission the `tests` subdir with its own `Makefile` was provided by @DerDakon. Thanks :-)

DerDakon:
@schmonz agreed that I can add another testcase I was just playing with, so I have extended this by a second commit that tests additional behavior of the stralloc functions.

schmonz:
Happy with the additions, at a glance. Please feel free to merge.

PR #103, opened 2019-09-12
remove functions that only reimplement standard C functions

DerDakon:
To the best of my knowledge all of these functions behave identical to their standard counterparts, so just use them instead.

In contrast to #65 this does _not_ remove ```byte_zero()``` as that is used to clear passwords in memory. A ```memset()``` could be optimized out by the compiler as the value is not used afterwards anymore, so keeping that function has a slightly higher chance of getting things cleared. One could check for the existence of ```explicit_bzero()``` or ```memset_s()``` and use that instead, but I think it overcomplicates things and just keeping it does no immediate harm. If we ever go to kill qmail-pop3d and qmail-popup we can go with a plain memset() then.


leahneukirchen:
Looks all good, str_len now returns a size_t which generally is larger than unsigned int, but this should be truncated to an int fine (this will be another incoming patch...).

qmail-local.c
```diff
@@ -388,7 +388,7 @@ int *cutable;
   if (qmeexists(fd,cutable)) {
     if (safeext.len >= 7) {
       i = safeext.len - 7;
-      if (!byte_diff("default",7,safeext.s + i))
+      if (memcmp("default",safeext.s + i,7) == 0)
```
josuah, line 391: Why replacing it inline specifically here? Because it was the only occurrence?

This raises my interest on semantics and guidelines to keep for the project:
This means we aim to replace them inline when possible, unless where too many parts would change (and put more chance to break patches)?

DerDakon: Because it should have been ```byte_equal()``` from the start. This is the only ```byte_diff()``` so I thought I could replace it with the final value if I touch it anyway.

tests/unittest_byte.c
```diff
@@ -0,0 +1,62 @@
+#include <check.h>
+
+#include "byte.h"
+
+START_TEST(test_byte_equal)
+{
+  ck_assert_int_ne(byte_equal("", 0, ""), 0);
```
schmonz: I might prefer `ck_assert_int_eq(..., 1)` if we were keeping these tests around. I’m really happy they’re here for just long enough to assure the transition to libc calls. Thank you for writing them.

tests/Makefile
```diff
@@ -9,7 +9,7 @@ default: it
 
 .PHONY: clean default it test
 
-TESTBINS = unittest_stralloc
+TESTBINS = unittest_byte unittest_str unittest_stralloc
```
schmonz: Would be good to put this commit right after the .gitignore one (possibly squashing), so the tests are in place before changing anything.

Makefile
```diff
@@ -1775,10 +1775,10 @@ scan.h fmt.h
 	./compile splogger.c
 
 str.a: \
-makelib str_len.o str_diff.o str_diffn.o str_cpy.o str_chr.o \
+makelib str_diff.o str_diffn.o str_cpy.o str_chr.o \
 str_rchr.o str_start.o byte_chr.o byte_rchr.o byte_diff.o byte_copy.o \
 byte_cr.o byte_zero.o
-	./makelib str.a str_len.o str_diff.o str_diffn.o str_cpy.o \
+	./makelib str.a str_diff.o str_diffn.o str_cpy.o \
```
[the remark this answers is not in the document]
DerDakon: It will then also use ```str.h```, where it is defined to be ```strlen()```, which is in libc, so everything is fine.

tests/unittest_byte.c
DerDakon: The return value is only used to be checked for being ```if (byte_equal(…))``` or ```if (!byte_equal(…))```, so it's only important if it is 0 or not. In fact, the return value is the difference between the 2 characters in the string (e.g. ```'o' - 'e'```), but we don't care. And we can't and need not to make sure that this is the same return value in the libc implementation.

tests/Makefile
DerDakon: ignore what GitHub says about the order, look at git log.

tests/unittest_byte.c
schmonz: Aha! Thank you for this explanation.

TARGETS
```diff
@@ -91,16 +91,12 @@ error.o
 error_str.o
 error_temp.o
 error.a
-str_len.o
-str_diff.o
-str_diffn.o
 str_cpy.o
```
schmonz, line 94: This one too

str.h
```diff
@@ -1,14 +1,16 @@
...
+#define str_copy(s,t) strcpy((s),(t))
+#define str_diff(s,t) strcmp((s),(t))
+#define str_diffn(s,t,len) strncmp((s),(t),(len))
+#define str_len(s) strlen((s))
 extern unsigned int str_chr();
 extern unsigned int str_rchr();
 extern int str_start();
 
-#define str_equal(s,t) (!str_diff((s),(t)))
+#define str_equal(s,t) (strcmp((s),(t)) == 0)
```
schmonz, line 14: Simple enough to skip testing?

DerDakon: Yes please.

PR #105, opened 2019-10-05
use POSIX compliant calling of head and tail

DerDakon:


DerDakon:
There are older versions of coreutils that did not support the -1 way, but this has been readded long ago. Gentoo has a special fixup function in ebuilds to change things to the -n version, and the Linux kernel has IIRC fixed also the usage of it to be compliant. So if something is not really old and rotten I guess it will just work as it should.

leahneukirchen:
LGTM.

PR #106, opened 2019-10-15
get rid of "compile" command

DerDakon:
Let make handle all this. This also allows that one can directly call

  make CC=something CFLAGS="-O2 -Wall"

This needs a "recent" version of GNU make (>= 4) or BSD make so the ```!=``` syntax is supported. No idea what current MacOS ships.


DerDakon:
You can build individual object files or targets with different CC flags as you like and are not forced to used the same ones on everything.

josuah:
> Hmmm. The existing approach has the nice property of making an explicit dependency of the compile-time flags.

This sounds like a drawback for developers, and eventually packagers doing "tweak flags, recompile, tweak flags, recompile".

A workaround would be having a script that flush the flags to a file only if the file content differs, and use this file as a dependency of everything (like ./compile now).  The effect is changing the mtime if the flags have changed, thus rebuilding everything in that case.

So in case we want to upgrade the build system while still keeping the correctness of the current one, it might be possible. Although that would require an extra script and could break patches.

DerDakon:
You can use conf-cc as always, and nothing will get lost. The Makefile will just _additionally_ behave like every other Makefile out there and allow direct override. I doubt any packager would fall in a rebuild trap there, they usually start with a clean source tree for reproducability anyway.

DerDakon:
My intention is actually the other way round: let make call CC directly so I can see the commands in the log afterwards. compile get's deleted, but the log is usually still there.

PR #107, opened 2019-10-17
Add configure script

### LINKED ISSUE #9: Make it easy to configure (the build), build, and install
It would be great to make it easy to configure (the build), build, and install notqmail.  Doing all that by hand is a pain right now.  Amitai Schleier has done an excellent job of automating all of that in [pkgsrc](https://pkgsrc.org/) with the [qmail-run](https://schmonz.com/qmail/pkgsrc-qmail-run/) package, but obviously that requires using pkgsrc, and I worry that that is a big enough barrier to entry for some people that they won't consider notqmail.

Even though I'm not a big fan of Autotools, being able to do `./configure && make && make install` is pretty convenient, and a lot of package management systems support it.  If not actually Autotools, it could be an Autotools workalike.  (Or something else, I suppose, even though not as standard as the Autotools interface, assuming it requires running just a few commands, would really be a big improvement.)

This would also increase the chance of notqmail being included in other OS package repositories (e.g., Homebrew, RHEL, Ubuntu, Arch, Gentoo, Alpine, etc.) which would also be a big plus.  It's certainly convenient for someone intending a production deployment, but it's also convenient for someone wishing to try it out.

Lastly, this would make it easier for someone writing a tutorial.  There are a number of tutorials on the Internet on how to install and configure software X on operating system Y.  For example, DigitalOcean seems to host a number of these (not surprisingly since I assume they hope people will use their services for hosting), and a Google search (e.g., `how install postfix rhel`) will often find them.  When a tutorial author can write, e.g., "First, install Postfix: `# yum install postfix`," that's really easy compared to what they'd have to write without a binary package being available in the OS's package repository.

DerDakon:
Fixes #9

### DIFF
--- a/configure
+++ b/configure
@@ -1,0 +1,58 @@
+#!/bin/sh
+
+AUTO_QMAIL=/var/qmail
+
+for arg in $*; do
+	case "${arg}" in
+	--prefix=*)
+		AUTO_QMAIL=$(echo "${arg}" | sed 's#^--prefix=##')
+		;;
+	--help)
+		echo "--prefix    install prefix [/var/qmail]"
+		exit 0
+		;;
+	*)
+		echo "unknown argument: ${arg}" >&2
+		exit 1
+		;;
+	esac
+done
+
+[ -z "${CC}" ] && CC=cc
+[ -z "${CFLAGS}" ] && CFLAGS=-O2
+[ -z "${LD}" ] && LD="cc -s"
+[ -z "${AR}" ] && AR=ar
+[ -z "${RANLIB}" ] && RANLIB=ranlib
+
+cat > conf-cc << EOF
+${CC} ${CFLAGS}
+
+This will be used to compile .c files.
+EOF
+
+cat > conf-ld << EOF
+${LD}
+
+This will be used to link .o files into an executable.
+EOF
+
+cat > conf-qmail << EOF
+${AUTO_QMAIL}
+
+This is the qmail home directory. It must be a local directory, not
+shared among machines. This is where qmail queues all mail messages.
+
+The queue (except for bounce message contents) is crashproof, if the
+filesystem guarantees that single-byte writes are atomic and that
+directory operations are synchronous. These guarantees are provided by
+fixed-block filesystems such as UFS and by journaling filesystems. Under
+Linux, make sure that all mail-handling filesystems are mounted with
+synchronous metadata.
+EOF
+
+cat > make-makelib.sh << EOF
+echo 'main="\$1"; shift'
+echo 'rm -f "\$main"'
+echo '${AR} cr "\$main" \${1+"\$@"}'
+echo '${RANLIB} "\$main"'
+EOF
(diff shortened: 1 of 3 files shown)


configure
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
+
+AUTO_QMAIL=/var/qmail
+
+for arg in $*; do
+	case "${arg}" in
+	--prefix=*)
+		AUTO_QMAIL=$(echo "${arg}" | sed 's#^--prefix=##')
```
josuah, line 8: is AUTO_QMAIL=${arg#*=} to avoid instead of the more explicit AUTO_QMAIL=$(echo "${arg}" | sed 's#^--prefix=##')? This is a minor detail, I am fine with both styles.

configure
```diff
@@ -0,0 +1,58 @@
...
+		;;
+	--help)
+		echo "--prefix    install prefix [/var/qmail]"
```
josuah, line 11: if we document `--prefix` and not `--prefix=dir`, we would have to support `--prefix dir`?

configure
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
+
+AUTO_QMAIL=/var/qmail
+
+for arg in $*; do
```
josuah, line 5: I suggest using `for arg; do` which does the same as `for arg in "$@"; do`, which helps in case any argument has a space in it, which is not expected anyway, so I am ok with `for arg in $*; do` nontheless.

configure
```diff
@@ -0,0 +1,58 @@
...
+fixed-block filesystems such as UFS and by journaling filesystems. Under
+Linux, make sure that all mail-handling filesystems are mounted with
+synchronous metadata.
+EOF
+
+cat > make-makelib.sh << EOF
+echo 'main="\$1"; shift'
+echo 'rm -f "\$main"'
+echo '${AR} cr "\$main" \${1+"\$@"}'
+echo '${RANLIB} "\$main"'
```
josuah, line 57: Now, we have a script (`./configure`) that generates a script (`make-makelib.sh`) that is used to generate the script (`make-makelib`) that generates the script (`makelib`) that generates the library.

I will propose an extra branch for changing the make-makelib.sh to make AR and RANLIB configurable like CC and LD.

configure
DerDakon: That is a bash extension.

configure
DerDakon: good catch

configure
josuah: It does not look like so, but the sed version is also very clear and good.

> `${parameter#[word]}`
>   Remove Smallest Prefix Pattern. The word shall be expanded to produce a pattern. The parameter expansion shall then result in parameter, with the smallest portion of the prefix matched by the pattern deleted. If present, word shall not begin with an unquoted '#'.
> https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

josuah:
Note that this branch could have to be re-written for #35, which fits nicely to the model you have chosen for ./configure.

Issue #83, opened 2019-08-24
is the variable firstwhen used uninitialized in tcpto.c?


alanpost:
If one were to compile the qmail code with -Wall -Werror, the following error is produced:

    ./compile tcpto.c
    tcpto.c: In function ‘tcpto_err’:
    tcpto.c:139:34: error: ‘firstwhen’ may be used uninitialized in this
    function [-Werror=maybe-uninitialized]
          if ((firstpos < 0) || (when < firstwhen))
                                ~~~~~~^~~~~~~~~~~~

Is there a condition where `firstwhen` is actually used uninitialized, or is the compiler issuing a spurious report about a condition that cannot in truth happen?  If someone is interested in tracing this code out and reporting back, this investigation makes a good first issue.

mbhangui:
firstwhen is always -1 before the for (i = 0; i< n; ++) loop. Hence the condition (when < firstwhen) never gets evaulated in the first iteration of the for loop. In indimail, I have initiliazed firstwhen to zero to take care of the compiler warning.

mbhangui:
It looks like Dan just wanted to give us a hard time understanding that line.

schmonz:
@mbhangui, since you've worked this through, can you submit a PR with your fix? Should be a quick review and merge.

mbhangui:
I have created a pull request #204 for this

PR #111, opened 2019-12-27
move some variables to a more local scope

DerDakon:
No target release given, I would just merge it once positive reviews are there.


DerDakon:
I was just looking again on #83.

PR #113, opened 2019-12-31
rename local variables shadowing global variables of the same name

DerDakon:
Again I would merge this whenever it is ready, preferable even for 1.08.


DerDakon:
Yes, running with -Wshadow.

DerDakon:
The part in cdbmake_add.c collides with #124. It can just be dropped when that is merged first.

PR #117, opened 2020-01-08
make the queue directory world readable

DerDakon:
This affects only the queue directory (i.e. /var/qmail/queue) itself, but not any of the subdirectories. The idea is that one can open that directory and then is able to call e.g. fstatvfs() on it. This is e.g. used by Qsmtp to check if a given mail size (by SIZE extension) is too large for the queue and avoid needless network traffic.

### DIFF
--- a/hier.c
+++ b/hier.c
@@ -49,7 +49,7 @@ void hier()
 
   d(auto_qmail,"alias",auto_uida,auto_gidq,02755);
 
-  d(auto_qmail,"queue",auto_uidq,auto_gidq,0750);
+  d(auto_qmail,"queue",auto_uidq,auto_gidq,0755);
   d(auto_qmail,"queue/pid",auto_uidq,auto_gidq,0700);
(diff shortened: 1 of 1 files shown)


DerDakon:
Just to make this explicit: any queue contents are still invisible to anyone not in group qmail as all subdirectories of queue/ have "other" permissions "---".

josuah:
I doubt DJB had use of fstatvfs(), which lead him to have the queue directory readable only by the owner or group members, and not to prevent stat(2)-ing the content orereaddir(3) them out, but rather because there was no reason to let anyone have access to it at that time.

On the other hand, it becomes really necessary to ensure correct permissions on the subdirectories.

Having a tool to recreate the queue tree with correct permissions if needed (#61) would make handling correct permissions safer and improve the situation after this change.

DerDakon:
If one had said, "hey, welcome to the 21st century, we have O_PATH for a while", that would have been cool. Long story short: with O_PATH on Linux the statvfs() call can be done on the queue directory even without it being readable.

PR #118, opened 2020-02-03
Add acknowledgement for the bug fix contribution to qmail-local.c

eriksjolund:
Reference:
https://web.archive.org/web/20060411195013/http://www.ornl.gov/lists/mailing-lists/qmail/2000/10/msg00696.html

Signed-off-by: Erik Sjölund <erik.sjolund@gmail.com>

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -40,7 +40,8 @@
 20031027 code: qmail-smtpd identifies itself as netqmail
 20031027 code: if $QMAILQUEUE is set, it's invoked instead of qmail-queue
 20031024 code: changed errno from int to #include.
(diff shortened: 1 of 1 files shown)


DerDakon:
In general: yes. I'm just unsure if this belongs at that place in the README, because the actual contribution was to nEtqmail.

DerDakon:
If that way is fine with you: I'm all for it. Please just squash these 2 commits together.

eriksjolund:
It looks good to me. Thanks

schmonz:
Since we seem to be the git repo of record, I'm all for this. Thanks, @eriksjolund!

PR #121, opened 2020-04-08
qmail-remote: remove flagcname, deprecated feature

### LINKED ISSUE #120: qmail-remote: remove CNAME handling in addrmangle
Currently, when qmail-remote sends a mail (not to a RELAYHOST) to `foo@bar.org`, and `bar.org` is a CNAME to `quux.org`, addrmangle will rewrite the envelope address to `foo@quux.org`. Aeons ago, this was deemed a good idea. Nowadays, at least Postfix (since version 2 from 2002) and OpenSMTPD (never) don't do it. Sendmail can do it, but apparently doesn't by default.

RFC 5321 (from 2008), Section 5.1 clearly specifies what MTA should do today:

>   The lookup first attempts to locate an MX record associated with the
>  name.  If a CNAME record is found, the resulting name is processed as
>   if it were the initial name.

djb wrote in https://cr.yp.to/im/cname.html:

> Most implementors agree that this feature is not worth the implementation costs. I recommend that, before January 2001, sites stop relying on client-side CNAME translation, so that implementors can safely remove the feature at that time.

It is easy to patch out by considering `flagcname` to be constant zero.

leahneukirchen:
qmail-remote used to rewrite the RCPT TO: domain when it was a CNAME,
as it was common before 2001.

RFC 5321 (from 2008), Section 5.1 clearly specifies what MTA should do
today:

    The lookup first attempts to locate an MX record associated with the
    name. If a CNAME record is found, the resulting name is processed as
    if it were the initial name.

djb wrote in https://cr.yp.to/im/cname.html:

    Most implementors agree that this feature is not worth the
    implementation costs. I recommend that, before January 2001, sites stop
    relying on client-side CNAME translation, so that implementors can
    safely remove the feature at that time.

Fixes #120.

### DIFF
--- a/qmail-remote.c
+++ b/qmail-remote.c
@@ -275,14 +275,9 @@ void smtp()
 stralloc canonhost = {0};
 stralloc canonbox = {0};
 
-void addrmangle(saout,s,flagalias)
-stralloc *saout; /* host has to be canonical, box has to be quoted */
-char *s;
-int *flagalias;
+void addrmangle(stralloc *saout, char *s)
 {
   int j;
- 
-  *flagalias = 0;
  
   j = str_rchr(s,'@');
   if (!s[j]) {
@@ -291,6 +286,7 @@ int *flagalias;
   }
   if (!stralloc_copys(&canonbox,s)) temp_nomem();
   canonbox.len = j;
+  /* box has to be quoted */
   if (!quote(saout,&canonbox)) temp_nomem();
   if (!stralloc_cats(saout,"@")) temp_nomem();
  
@@ -324,8 +320,6 @@ int main(int argc, char **argv)
   unsigned long random;
   char **recips;
   unsigned long prefme;
-  int flagallaliases;
-  int flagalias;
   char *relayhost;
  
   sig_pipeignore();
@@ -353,18 +347,16 @@ int main(int argc, char **argv)
   }
 
 
-  addrmangle(&sender,argv[2],&flagalias);
+  addrmangle(&sender,argv[2]);
  
   if (!saa_readyplus(&reciplist,0)) temp_nomem();
   if (ipme_init() != 1) temp_oserr();
  
-  flagallaliases = 1;
   recips = argv + 3;
   while (*recips) {
     if (!saa_readyplus(&reciplist,1)) temp_nomem();
     reciplist.sa[reciplist.len] = sauninit;
-    addrmangle(reciplist.sa + reciplist.len,*recips,&flagalias);
-    if (!flagalias) flagallaliases = 0;
+    addrmangle(reciplist.sa + reciplist.len,*recips);
     ++reciplist.len;
     ++recips;
   }
@@ -388,7 +380,6 @@ int main(int argc, char **argv)
         prefme = ip.ix[i].pref;
  
   if (relayhost) prefme = 300000;
-  if (flagallaliases) prefme = 500000;
  
   for (i = 0;i < ip.len;++i)
     if (ip.ix[i].pref < prefme)
--- a/qmail-remote.c
+++ b/qmail-remote.c
@@ -275,15 +275,14 @@ void smtp()
 stralloc canonhost = {0};
 stralloc canonbox = {0};
 
-void addrmangle(saout,s,flagalias,flagcname)
+void addrmangle(saout,s,flagalias)
 stralloc *saout; /* host has to be canonical, box has to be quoted */
 char *s;
 int *flagalias;
-int flagcname;
 {
   int j;
  
-  *flagalias = flagcname;
+  *flagalias = 0;
  
   j = str_rchr(s,'@');
   if (!s[j]) {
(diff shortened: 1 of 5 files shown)


DerDakon:
At first glance looks good. If you change the interface of a function please convert it to C89 style.

leahneukirchen:
If someone knows about the idea behind `if (flagallaliases) prefme = 500000;` i'd be curious to learn about it.

DerDakon:
In general I think this is correct. I have not really looked into the code, i.e. what it is actually doing, and I don't want this in 1.08 if we don't agree on that on a broader scope. So, no +1 for me for the moment, but it will likely happen.

leahneukirchen:
Sure, this has been wrong since 2008 so one release more or less doesn't matter. ;)

josuah:
> If someone knows about the idea behind `if (flagallaliases) prefme = 500000;` i'd be curious to learn about it.

`prefme` is compared with `.pref` from that `ipalloc` struct (an array of `struct { ip; pref; }`).

### What is .pref and where is it taking its data from?

In `dns.c:int dns_mxip(ipalloc *ia, stralloc *sa, unsigned long random);`, we have a call to `findmx()`, iterated to scan all the MX records found from DNS.

The position, preference and IP of the MX record are passed through static global variables to `dns_mxip()` (why not... no threads).

For each result, `ia` is filled with the IPs one by one through `static int dns_ipplus(ipalloc *ia, stralloc *sa, int pref)`, and that is where `qmail-remote` gets its list of preferences: from DNS only, no configuration impacting `ip[].ix.pref` at all.

### What is the effect of comparing prefme with each ip[].ix.pref from DNS?

Just below `if (flagallaliases) prefme = 500000;`, it chooses the first ip.[].ix.pref ip that is lower than the `prefme` threshold.

Just above `if (flagallaliases) prefme = 500000;`, the loop finds the lowest pref value for IPs matching those of the server (SIOCGIFCONF) and store it into `pref`.

That filters out all the MX entries that have a lower preference than qmail's own IPs.  That looks like having the effect of capturing the mail in case qmail-remote has to send it out, but the server it send it to has the same IP (in other words: send a mail to myself through IP).

It is still possible to send a mail to another mail server among the MX entries, even if there is an IP of the local interface in the pool of MX responses : with a lower MX than the one that match our IPs.

This looks like a safeguard against misconfiguration: a mail for a same IP as one of which qmail listen on needs to be send through `qmail-local`, not through `qmail-remote`!

### Why a high value for prefme in some cases?

When the mail server needs to relay everything to somewhere (special case), or for these addrmangle cases (tricky cases), `qmail-remote` bypasses this mechanism: allow all ips regardless of the context, and gonna do what it gotta do what it gotta do: forward to the first of the IPs found.

### Why 500000, why not 5725 or 42?

No idea! Debugging purposes? https://tools.ietf.org/html/rfc1035#section-3.3.9 says int16, so it maxes out to 65536.

A `printf()` (sorry, `buffer_putn()` :P) after the two loop would show whether the mechanism was triggered or not.

### A note on anycast BGP

In some situations, a public address might be at the same time on the interface onto which qmail listens, and another remote host, all that legitimately: Anycast BGP, for which the same IP is announced at multiple places, and the routing will make it go to the closest available.

The safeguard mechanism will still match the semantics of anycast BGP: if it reached any of that host with the one IP declared in DNS, the mail is home.  If there is something more to do (like backup to master mx transfer), a _lower_ priority has can be setup, and the mail will never be stuck on a relay, even if the same IP is on multiple hosts ("weird setups").

leahneukirchen:
The purpose of the prefme loop is to always forward mail to a higher-priority (= lower number) MX; if we are a relayhost, any MX is good.

But why the special case for flagallaliases?

josuah:
I am not sure.

I imagine it is so that it is actually sent out, even if it is to the same server, so that it goes through the whole chain with the new domain name, in case it behaves differently

DerDakon:
I'm not sure either, as from all I can see there can never be a pref > 0xffff. It is set in dns_mxip() to the 16 bit value from the DNS, or in dns_ip() to 0. It feels a bit like body-stripping, let it rest in pieces.

DerDakon:
I have looked around a bit, and not even the SPF patch uses ```dns_cname()```, so I guess there is no use for it anymore.

This also needs a changelog entry and the corresponding sections from the FAQ should be removed.

mbhangui:
On Fri, 1 May 2020 at 16:10, Rolf Eike Beer <notifications@github.com>
wrote:

> I have looked around a bit, and not even the SPF patch uses dns_cname(),
> so I guess there is no use for it anymore.
>
> This also needs a changelog entry and the corresponding sections from the
> FAQ should be removed.
>
There is one more target in Makefile - dnscname.c. It uses dns_cname() and
it gets compiled and linked but doesn't get installed as it is not in
hier.c. So this code is just lying there without any real use. This too
needs to be removed.

leahneukirchen:
This is already gone in master, I think:
CHANGES:20190822 cleanup: remove unused dnscname.

123fa007ca4e6e5ba55c62775c5d24a6c3ab794d

DerDakon:
From my feeling the last to patches should be squashed into the first one, which actually removed that feature. The other ones could optionally be squashed in or just kept as they are.

qmail-remote.c
```diff
@@ -276,34 +276,22 @@ void smtp()
...
-  *flagalias = flagcname;
- 
   j = str_rchr(s,'@');
   if (!s[j]) {
     if (!stralloc_copys(saout,s)) temp_nomem();
     return;
   }
   if (!stralloc_copys(&canonbox,s)) temp_nomem();
   canonbox.len = j;
+  /* box has to be quoted */
```
schmonz, line 290: Seems clear enough already from the following line.

schmonz: Oh, I see now this was the still-relevant part of the comment from old line 279. I’m okay keeping or getting rid of it, either way.

Issue #28, opened 2019-07-15
Investigate generating Makefile


schmonz:
Hand-editing `Makefile` is error-prone. DJB probably used [his redo](https://cr.yp.to/redo.html). Since that's not available to us, maybe [Avery Pennarun's redo](https://github.com/apenwarr/redo) would help?

xenotrope:
I experimented with that version of redo once and IIRC it runs mostly by the power of dark mojo Python libraries I couldn't comprehend. I _think_ he briefly mentions that he'll one day rewrite it in a language like C.

I was more satisfied with JdBP's redo: https://jdebp.eu/Softwares/redo/ but I found it it, too, lacking in some of the basic building blocks that I'd want from an out-of-the-box development environment. Case in point, a number of build scripts in djb are automatically created at compile time because the Makefile has things like:

compile: \
conf-cc print-cc.sh systype warn-auto.sh
        rm -f compile
        sh print-cc.sh > compile
        chmod 755 compile

redo doesn't rely on make or GNU make, so it doesn't naturally allow for one to hack this kind of wrapper-script-for-a-tool into it as easily. I wound up having to cobble this kind of thing into a custom default.do file based on what special actions to take given the input target filename:

#!/bin/sh
if [ -r $1=x ]; then
  redo-ifchange load
  redo-ifchange $1.o
  ./load $3 $1.o
  exit 0
fi

case $1 in
  compile)
    redo-ifchange conf-cc
    redo-ifchange print-cc.sh
    rm -f compile
    sh print-cc.sh > $3
    chmod 0755 $3
    exit 0
    ;;
[et cetera]

This is sloppy, but not much sloppier than the original Makefile. The best solution I found for a turnkey build environment was William Baxter's build-0.66, which utilized his djblibs-0.32 package to create slashpackage-friendly, automatic Makefile-making skeleton directories which would get you pretty productive pretty quickly. Need to print text in a djb-y way? Put a dozen or so buffer_*.o library dependency lines into ./src/unix=l and the "build" binary would automatically copy them into your project at compile time. I still use it to this day.

Unfortunately, Baxter replaced build-0.66 with a retooled 0.72 version that was missing a lot of the features 0.66 provided. I think he ended up taking both of those packages off of his website, but of course I still have local copies and the Internet Wayback Machine has an archive of his software site before he turned it into a consulting splash page.

https://web.archive.org/web/20141008040901/http://www.superscript.com/build/index.html

DerDakon:
[...]

PR #122, opened 2020-04-11
Rework the make-* scripts for simpler toolchain

josuah:
Each commit address an independent point, with the global goal of making the toolchain simpler, and easier to rework with further on: #28, #2, #35, #107


Makefile
```diff
@@ -908,25 +899,19 @@ warn-auto.sh mailsubj.sh conf-qmail conf-break conf-split
...
-make-compile: \
-make-compile.sh auto-ccld.sh
-	cat auto-ccld.sh make-compile.sh > make-compile
-	chmod 755 make-compile
-
-make-load: \
-make-load.sh auto-ccld.sh
-	cat auto-ccld.sh make-load.sh > make-load
-	chmod 755 make-load
+compile: \
```
DerDakon, line 902: Please keep the alphabetical order of the Makefile targets.

elq.sh
```diff
@@ -1 +1 @@
-QMAIL/bin/maildir2mbox && exec elm ${1+"$@"}
```
DerDakon, line 1: Don't touch elq, it will die right after 1.08. Same for pinq and qail.

makelib.sh
```diff
@@ -0,0 +1,4 @@
+main="$1"; shift
```
DerDakon, line 1: "main" isn't really needed, one could use $1 at the 2 places where it matters below and just $* for the AR line.

josuah: I replicated the previous scripts directly, I prefer as you suggest.

Makefile
josuah: I am now learning this best practice for large makefiles.

elq.sh
josuah: I applied the change to all except this one.

josuah:
FYI: Just rebased it to master.

PR #124, opened 2020-04-13
remove the custom allocator

### LINKED ISSUE #112: Remove the custom allocator (alloc.c)
What does this do at all? It has a 4096 byte static allocation and serves small memory blocks from there as long as they fit, probably to avoid crappy slow system allocators. I say: kill it. Here are some things I have in mind why we should do it:

 * it's as simple as possible, probably intentional. This makes things like ```alloc_re()``` (i.e. ```realloc()```) inefficient as it does not look if the thing it reallocates is already the last thing in the batch, and simply extends that, or if ```alloc_free()``` has the last block and adds things back to the pool (ideally with zeroing for security reasons).
 * it makes it basically impossible to detect overruns on early small allocations with tools like valgrind, address sanitizer or friends as those allocations never hit ```malloc()```, so are never traced by any custom tools. Any overrun will just go into the next variable, untracable.
 * allocators do their own buffering, alignment, and so on. If they are broken, your system is severely screwed. And if we need more money^H^H^H^Hemory we will hit that buggy allocatator anyway, making things harder to detect as it will only happen on "heavy" memory usage. This is not TruOS, SunOS 1.x or something like that anymore. We expect a reasonably sane OS. Everything else will get us screwed in much more severe ways.
 * it has shown issues when coming close to 4GiB allocations (#37, #109, [older reports](https://marc.info/?l=qmail&m=125213850310173&w=2)). DJB correctly says that an allocation of that size in your mailer shows that you have already done something wrong, but still: 64 bit platforms are the default, so using a 32 bit type to pass allocation sizes is a bug of it's own.
 * why bother at all with writing an allocator?

DerDakon:
* it's as simple as possible, probably intentional. This makes things like alloc_re() (i.e. realloc()) inefficient as it does not look if the thing it reallocates is already the last thing in the batch, and simply extends that, or if alloc_free() has the last block and adds things back to the pool (ideally with zeroing for security reasons).
 * it makes it basically impossible to detect overruns on early small allocations with tools like valgrind, address sanitizer or friends as those allocations never hit malloc(), so are never traced by any custom tools. Any overrun will just go into the next variable, untracable.
 * allocators do their own buffering, alignment, and so on. If they are broken, your system is severely screwed. And if we need more money^H^H^H^Hemory we will hit that buggy allocator anyway, making things harder to detect as it will only happen on "heavy" memory usage. This is not TruOS, SunOS 1.x or something like that anymore. We expect a reasonably sane OS. Everything else will get us screwed in much more severe ways.
 * it has shown issues when coming close to 4GiB allocations (#37, #109, older reports). DJB correctly says that an allocation of that size in your mailer shows that you have already done something wrong, but still: 64 bit platforms are the default, so using a 32 bit type to pass allocation sizes is a bug of it's own.
 * when any additional patch accidentially mixes calls to this functions with normal realloc()/free() this can lead to random crashes
 * why bother at all with having a custom allocator?

Fixes #112


alloc.h
```diff
@@ -1,8 +1,17 @@
...
 #define ALLOC_H
 
-extern /*@null@*//*@out@*/char *alloc();
-extern void alloc_free();
-extern int alloc_re();
+#include <stdlib.h>
+
+#define alloc(n) malloc(n)
+#define alloc_free(x) free(x)
+static inline int alloc_re(char **x, unsigned int m, unsigned int n)
```
josuah, line 8: I missed that `x` is a pointer passed as a reference, so it did segfault on my iteration, so good catch.

cdbmss.c
```diff
@@ -34,7 +34,7 @@ unsigned int datalen;
   for (i = 0;i < keylen;++i)
     h = cdbmake_hashadd(h,(unsigned int) key[i]);
 
-  if (!cdbmake_add(&c->cdbm,h,c->pos,alloc)) return -1;
+  if (!cdbmake_add(&c->cdbm,h,c->pos,malloc)) return -1;
```
josuah, line 37: In case we want to keep the source entirely untouched, we may use
```C
#define alloc malloc
``` 
But I am also happy to remove residues of `alloc` when possible.

alloc.h
DerDakon: You are not the only one ;)

cdbmss.c
schmonz: +1 to keep referring to `alloc` here, since we can.

schmonz:
I find the motivation for doing this very strong. The sooner we can turn on Valgrind and ASan in CI, the sooner we can be systematically ratcheting safety upward.

cdbmss.c
DerDakon: We would add a #define for only one line of code that no patch will likely ever touch. I don't see a benefit in this.

DerDakon: done

DerDakon:
We have 2 possibilities: either merge #113 first, which renames the usages of alloc in cdb code, or we just remove it. Either way, we have to touch that anyway as otherwise the ```alloc``` there will be changed from the define to ```malloc```, which would also affect the function parameter. IMHO the best thing is to just remove that entirely and use ```malloc()``` directly as there is no need to pass that as function pointer here. Afterwards the ```alloc(x)``` version of the define can be used again.

PR #126, opened 2020-04-21
fmt.h prototypes

leahneukirchen:
First steps towards a prototypification of the headers.

Should not cause merge conflicts with patch sets as these functions are usually untouched.


DerDakon:
I would reverse the order of the commits, I'm not sure if the compilers like it if a function has a C89 prototype and a K&R style declaration.

leahneukirchen:
That works fine, but I can also swap it.

DerDakon:
If it works it makes no difference and we can leave it as is.

PR #130, opened 2020-05-14
use default strdup() instead of open coding it

DerDakon:


DerDakon:
For the moment I would keep it as low as possible.

PR #132, opened 2020-05-19
Let NROFF be redefined in the environment.

schmonz:


DerDakon:
Can you tell me why this is needed? We have a testcase that overrides NROFF already and that works, so which case  is the problem?

schmonz:
Before:

```
:; env NROFF=definitely-not-a-command-available-on-my-system make man
nroff -man qmail-local.8 > qmail-local.0
[...]
```

After:

```
:; env NROFF=definitely-not-a-command-available-on-my-system make man
definitely-not-a-command-available-on-my-system -man qmail-lspawn.8 > qmail-lspawn.0
/bin/sh: definitely-not-a-command-available-on-my-system: command not found
make: *** [qmail-lspawn.0] Error 127
```

(And I'm not sure how we didn't notice the missing include in `instcheck.c` before now, but I'm glad it came up.)

DerDakon:
Ah, ok, I have only used the form with adds things as arguments to make, and our testcase does the same, so this is why that one slipped through.

DerDakon:
Could you change one of the CI runs (either Travis or Cirrus, but not both) to use the env form in the NROFF commit, too, so we can prove this works?

DerDakon:
The testcases are completely broken, it looks to me as the NROFF versions never actually did anything different than the normal ones.

Makefile
```diff
@@ -1,7 +1,7 @@
 # Don't edit Makefile! Use conf-* for configuration.
 
 SHELL=/bin/sh
-NROFF=nroff
+NROFF?=nroff
```
josuah, line 4: While `?=` is extremely convenient, it is not in [POSIX make](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html).

The semantics seems to be that `make -e` is for letting environment variables overriding make's variables, and letting the default not affect make.

My feel is that `-e` should be the default and assume that the user minds its environment variables.

Do we care? It looks like both BSDs makes and GNU make support `?=`.  I am fine with both ways.

```
# env NROFF=definitely-not-a-command-available-on-my-system make man make -e man
definitely-not-a-command-available-on-my-system -man splogger.8 > splogger.0
/bin/sh: definitely-not-a-command-available-on-my-system: not found
```

DerDakon: I was tempted to say "then let's do the more portable way", but we require a C99 capable compiler, so why not a at least basically sensible make.

schmonz: In my cross-platform pkgsrc travails I've not encountered a `make` without `?=`, and assumed it was standard. Interesting to hear it's not POSIX. I'm comfortable requiring it for the purpose of overriding `NROFF` :-)

PR #134, opened 2020-05-19
fix catman rules and tests

DerDakon:
Make the changes from #116 actually work as intended.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,19 @@ default: it
 
 .PHONY: check clean default it man test
 
-%.0:
-	$(NROFF) -man $^ > $@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -7,8 +7,10 @@ default: it
 
 .PHONY: check clean default it man test
 
-%.0:
-	$(NROFF) -man $^ > $@
+.SUFFIXES: .0 .1 .3 .5 .7 .8
+
+.1.0:
```
schmonz, line 12: same for sections 5, 7, 8?

DerDakon: obiously.

PR #141, opened 2020-05-23
Docker/Podman builds for notqmail

mbhangui:
# Dockerfile repository for automated builds.
  
This repo contains repository for Dockerfiles used for building
docker/podman images for notqmail

https://hub.docker.com/r/docker_username/notqmail

The following tags/images can be pulled by executing
the commands

## a) docker

```
docker pull docker_user/notqmail:tag
```

or

## b) podman

```
podman pull docker_user/notqmail:tag
```

Replace tag in the above command with one of the following
```
xenial     for Ubuntu 16.04
bionic     for Ubuntu 18.04
focal      for Ubuntu 20.04
centos7    for CentOS 7
centos8    for CentOS 8
debian8    for Debian 8
debian9    for Debian 9
debian10   for Debian 10
fc31       for Fedora Core 31
fc32       for Fedora Core 32
Tumbleweed for openSUSE Tumbleweed
Leap15.2   for openSUSE Leap 15.2
```

## Instructions for starting the docker/podman container
(replace podman with docker for docker operations)

### list podman images

$ podman images
```
REPOSITORY                           TAG       IMAGE ID       CREATED          SIZE
localhost/notqmail                   centos7   4bcbc7a876fe   38 minutes ago   452 MB
localhost/notqmail                   focal     d3b533841934   2 hours ago      141 MB
```

### Start the podman container

notqmail uses docker-entrypoint to execute qmail-start and start qmail-send, qmail-lspawn, qmail-rspawn.
You just need to pass any argument other than notqmail to bypass the default entrypoint.

$ podman run -d -h notqmail.org --name notqmail d3b533841934
```
08a4df5054d920cfdf8869aa777a7afc39bab19591394ea283c0c082f8b0a876
```

You can use --net host to map the container's network to the HOST
```
$ docker run --net host -d -h notqmail.org --name notqmail d3b533841934
or
$ podman run --net host -d -h notqmail.org --name notqmail d3b533841934
```

### Query the id of the container

$ podman ps
```
CONTAINER ID  IMAGE                                   COMMAND   CREATED             STATUS                 PORTS  NAMES
08a4df5054d9  docker.io/cprogrammer/notqmail:focal    notqmail  About a minute ago  Up About a minute ago         notqmail
```

### Execute an interactive shell in the container

$ podman exec -ti notqmail /bin/bash
```
root@notqmail:/#
```

### Get processlist in the container

```
root@notqmail:/# ps -ef
UID          PID    PPID  C STIME TTY          TIME CMD
qmails         1       0  0 10:10 pts/0    00:00:00 qmail-send
qmaill        35       1  0 10:10 pts/0    00:00:00 splogger qmail
root          36       1  0 10:10 pts/0    00:00:00 qmail-lspawn ./Maildir/
qmailr        37       1  0 10:10 pts/0    00:00:00 qmail-rspawn
qmailq        38       1  0 10:10 pts/0    00:00:00 qmail-clean
syslog        50       1  0 10:10 ?        00:00:00 /usr/sbin/rsyslogd
root          66       0 40 10:22 pts/1    00:00:00 bash
root          75      66  0 10:22 pts/1    00:00:00 ps -ef
root@notqmail:/#
```

### Get console logs of the container
```
$ podman logs
Warning  linking /etc/mtab to /proc/self/mounts
Your fully qualified host name is notqmail.org.
Putting notqmail.org into /var/qmail/control/me...
Putting notqmail.org into /var/qmail/control/defaulthost...
Putting notqmail.org into /var/qmail/control/envnoathost...
Putting notqmail.org into /var/qmail/control/defaultdomain...
Putting notqmail.org into /var/qmail/control/plusdomain...
Putting notqmail.org into /var/qmail/control/locals...
Putting notqmail.org into //var/qmail/control/rcpthosts...
Now qmail will refuse to accept SMTP messages except to the following...
notqmail.org
Make sure to change rcpthosts if you add hosts to locals or virtualdomains!
 * Starting enhanced syslogd rsyslogd
```

### Stop the container

$ podman stop \`podman ps -q\`
```
08a4df5054d920cfdf8869aa777a7afc39bab19591394ea283c0c082f8b0a876
```

### Clear the stopped container image

$ podman rm \`podman ps -aq\`
```
08a4df5054d920cfdf8869aa777a7afc39bab19591394ea283c0c082f8b0a876
```

## c) github respository for Dockerfile

The Dockerfile for each of the images is located in a separate subdirectory for each linux distro

**notqmail**
https://github.com/notqmail/docker/tree/master/notqmail

If you want to build the image yourself instead of using hub.docker.com,
you can use the below commands

```
$ docker build -t notqmail:focal ./Dockerfile.focal .
or
$ podman build -t notqmail:focal ./Dockerfile.focal .
```

## NOTE THIS FILE REQUIRES Editing after completing following steps ##
1. create a user on hub.docker.com (docker_user)
2. Create a repository for docker files and update point c) above
Take a look at

https://github.com/mbhangui/docker
&
https://hub.docker.com/repository/docker/cprogrammer/indimail-mta

### DIFF
--- a/docker/docker-entrypoint
+++ b/docker/docker-entrypoint
@@ -1,0 +1,37 @@
+#!/bin/sh
+# WARNING: This file was auto-generated. Do not edit!
+#
+set -e
+
+case "$1" in
+notqmail)
+	if [ ! -f /etc/mtab ] ; then
+		if [ -f /proc/self/mounts ] ; then
+			echo "Warning  linking /etc/mtab to /proc/self/mounts" 1>&2
+			ln -s /proc/self/mounts /etc/mtab
+		elif [ -f /proc/mounts ] ; then
+			echo "Warning  linking /etc/mtab to /proc/mounts" 1>&2
+			ln -s /proc/mounts /etc/mtab
+		else
+			echo "Warning /etc/mtab: No such file or directory" 1>&2
+		fi
+	fi
+	if [ ! -f /var/qmail/control/me ] ; then
+		HOSTNAME=`hostname`
+		echo $HOSTNAME > /var/qmail/control/me
+		/usr/libexec/notqmail/config-fast $HOSTNAME
+	fi
+	# this is a minimal system
+	# start system logger for splogger to log entries
+	if [ -x /usr/lib/systemd/systemd-journald ] ; then
+		/usr/lib/systemd/systemd-journald &
+	elif [ -f /etc/init.d/rsyslog ] ; then
+		/etc/init.d/rsyslog start
+	fi
+	exec env PATH=/var/qmail/bin:$PATH qmail-start ./Maildir/ splogger qmail
+;;
+*)
+	echo "docker-entrypoint: executing $@"
+	exec "$@"
+;;
+esac
--- a/docker/Dockerfile.CentOS7
+++ b/docker/Dockerfile.CentOS7
@@ -1,0 +1,36 @@
+FROM centos:7
+MAINTAINER cprogrammer
+
+# Freshen RUN yum update
+RUN mkdir -p /root /usr/libexec/notqmail \
+	&& yum -y install \
+		deltarpm \
+	&& yum -y update \
+	&& yum -y install \
+		gnupg2 \
+		binutils \
+		hostname \
+		mailx \
+		net-tools \
+		openssl \
+		procps \
+		psmisc \
+		telnet \
+		wget \
+		which \
+	&& wget \
+		https://download.opensuse.org/repositories/home:notqmail/CentOS_7/home:notqmail.repo \
+		-O /etc/yum.repos.d/notqmail.repo \
+	&& yum -y install notqmail \
+	&& (echo "/var/qmail/alias/Maildir/" > /var/qmail/alias/.qmail-root; \
+		mkdir -p /var/qmail/alias/Maildir/tmp; \
+		mkdir -p /var/qmail/alias/Maildir/new; \
+		mkdir -p /var/qmail/alias/Maildir/cur; \
+		chown -R alias:qmail /var/qmail/alias/Maildir) \
+	&& (alternatives --install /usr/sbin/sendmail mta /var/qmail/bin/sendmail 120 \
+        --slave /usr/bin/mailq mta-mailq /var/qmail/bin/qmail-qread \
+        --slave /usr/lib/sendmail mta-sendmail /var/qmail/bin/sendmail) \
+	&& ln -s /usr/libexec/notqmail/docker-entrypoint /usr/sbin/docker-entrypoint
+COPY docker-entrypoint config-fast /usr/libexec/notqmail
+ENTRYPOINT ["docker-entrypoint"]
+CMD ["notqmail"]
--- a/docker/Dockerfile.bionic
+++ b/docker/Dockerfile.bionic
@@ -1,0 +1,37 @@
+# docker build -f dockerfile.focal -t cprogrammer/indimail:focal
+# run dpkg-reconfigure tzdata after starting the container
+From ubuntu:bionic
+MAINTAINER cprogrammer
+
+# Freshen RUN apt-get -y update
+RUN export DEBIAN_FRONTEND=noninteractive \
+	&& mkdir -p /root /usr/libexec/notqmail \
+	&& apt-get -y update \
+	&& apt-get -y install wget gnupg2 apt-utils \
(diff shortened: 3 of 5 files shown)


docker/docker-entrypoint
```diff
@@ -0,0 +1,37 @@
...
+		if [ -f /proc/self/mounts ] ; then
+			echo "Warning  linking /etc/mtab to /proc/self/mounts" 1>&2
+			ln -s /proc/self/mounts /etc/mtab
+		elif [ -f /proc/mounts ] ; then
+			echo "Warning  linking /etc/mtab to /proc/mounts" 1>&2
+			ln -s /proc/mounts /etc/mtab
+		else
+			echo "Warning /etc/mtab: No such file or directory" 1>&2
+		fi
+	fi
```
josuah, line 18: simply curious here: is that a boilerplate to get a full filesystem or is it specific to notqmail?

mbhangui: Not specific to notqmail. I have a virtualdomains package which uses getmntent() call to find out the filesystem in which the user's Maildir is getting created (when adding a new user). It then creates maildirsize file for quota calculation. I guess the same can be done by using statvfs() call, but I have not yet investigated. This error came up (IIRC) on the ubuntu images. Most podman images are having the /etc/mtab file. The mounted filesystems are also in /proc/self/mounts. So I created this link so that user creation could go on without errors.

mbhangui: Once you create this link, getmntent() library call doesn't fail. I believe getmntent() should be modified to not look at a hardcoded /etc/mtab file.

josuah: Thank you for the insight. Maybe we adding a short comment like `# in case getmntent(3) is needed` before ?

getmntent(3):
> DESCRIPTION
>       These routines are used to access the filesystem description file
>       /etc/fstab and the mounted filesystem description file /etc/mtab.

josuah:
I am interested in knowing what is the motive for bringing Docker into the repository:

1. Spinning a container as part of unit tests from time to time.
2. Something that packager can use to produce binaries.
3. Make a notqmail-from-source runnable docker for users.
4. Simply provide the work for whoever want to have a Docker for notqmail.
5. All of the above.
6. Something else.

mbhangui:
> I am interested in knowing what is the motive for bringing Docker into the repository:
> 
> 1. Spinning a container as part of unit tests from time to time.

It can be used for testing. Once can put few scripts which sends out emails to some standard email address and a reply is sent back. But I doubt we could be anywhere near achieving that. 

> 2. Something that packager can use to produce binaries.
No. The opensuse build service is already doing that. The docker images are being built by exploting the availability of the binary. Theoretically one can also build a docker image by installing make, gcc, git. Pull the source code and install the binary. This could be a better way to test the compilation of notqmail on all the linux distros.

> 3. Make a notqmail-from-source runnable docker for users.

Yes.
> 4. Simply provide the work for whoever want to have a Docker for notqmail.

Docker / podman provides an easy way for a user to quickly run a notqmail instanace on most platforms. Theoretically the user can run a notqmail ubuntu instance on Windows too on a macbook too. Once a docker image is ready, the configuration will be identical for all users.

When you run a docker instance on your production server, you will not be modifiying any files on the server. The docker or podman instance is an isolated container that has everything needed to run notqmail in an isolated container. This gives the ability to the user to minimize changes to his/her existing server (the only change required is to install podman or docker using apt-get, dnf or yum).

The ultimate goal should be to have ucspi-tcp, daemontools also installed. Then one can have a complete mail server setup in a container. Users can simply (regardless of their OS), pull the docker image and run the container using the docker run or the podman run command (using the --net) command. and by doing that the user can have incoming SMTP and outgoing delivery from his server.

mbhangui:
Also once you have a deployable podman container for notqmail, one can go to the next step. Deploy [kubernates](https://kubernetes.io/). I haven't reached there yet. But I know people who have deployed one of my fork of qmail and made a scalable solution.

josuah:
Adding more directories regarding various build, containerization, integration with other platforms... could mislead those discovering notqmail into thinking that these became part of the build process.  Some project even have a [Docker file instead of a Makefile](https://github.com/ponchio/untrunc).

The convention seems to be that everything that is not part of the project itself starts with a dot, so the same could go for docker.

Or maybe an extra "contrib" directory, or maybe grouping it with the other continuous integration configs #148 could help with standing it clear that notqmail is not tied to GitHub or Docker.

Is there a requirement or a conventions for the location of docker images from projects?

josuah:
Reading #148 again, it looks like /docker aside to /tests will help keeping the ambiguity away ! :)

PR #143, opened 2020-05-24
Rework the remaining str_* functions

DerDakon:
This is not really intended to be merged as-is, it should be a point of discussion.


qmail-send.c
```diff
@@ -173,14 +173,14 @@ char *recip;
    if (str_equal(sender + i - 4,"-@[]"))
     {
      unsigned int j = byte_rchr(sender,i - 4,'@');
-     unsigned int k = str_rchr(recip,'@');
-     if (recip[k] && (j + 5 <= i))
+     char *at = strrchr(recip,'@');
+     if (at && (j + 5 <= i))
```
josuah, line 177: I had a moment of doubt, but indeed: `str_rchr(recip,'@') == '\0'` is the same as `strrchr(recip,'@') == NULL`.

str_chr.c.gnu
```diff
@@ -0,0 +1,9 @@
+#define _GNU_SOURCE
```
josuah, line 1: Testing on musl (`Linux t470s.z0.is 5.4.35-0-lts #1-Alpine SMP Thu, 23 Apr 2020 10:36:12 UTC x86_64 Linux`):

```
t470s:$ cp str_chr.c.gnu str_chr.c && make clean test 2>&1 | grep -m 1 '^real'
real    0m 2.32s
t470s:$ cp str_chr.c.impl str_chr.c && make clean test 2>&1 | grep -m 1 '^real'
real    0m 2.32s
t470s:$ cp str_chr.c.orig str_chr.c && make clean test 2>&1 | grep -m 1 '^real'
real    0m 2.33s
```

It did not look like the performance improvement was worth the extra check, but maybe the gap is bigger on other libcs.

[EDIT]: added the original implementation

josuah: Here is how I tested:

- Added `for (int i = 0; i < 200000; i++) { [...] }` around `ck_assert_uint_eq(str_rchr([...]))`
- Added `time` before `./$$tbin || exit 1 ;`
- Ran the above commands.

DerDakon: The main idea was not to get a performance improvement, but just to junk that code. Nice if it doesn't hurt anyway ;)

case_lowerb.c
```diff
@@ -1,8 +1,6 @@
 #include "case.h"
 
-void case_lowerb(s,len)
-char *s;
-unsigned int len;
+void case_lowerb(char *s, unsigned int len)
```
josuah, line 3: `unsigned int len` feels like a `size_t`, let's see if the `int` -> `size_t` commit id up or downstream...

josuah: It looks like there are multiple commits that convert `int` to `size_t` at multiple places. If this is something we want, would another branch be more appropriate or do we adjust source as we touch it?

byte_zero.c
```diff
@@ -1,13 +1,8 @@
 #include "byte.h"
 
-void byte_zero(s,n)
-char *s;
-register unsigned int n;
+void byte_zero(char *s, unsigned int n)
```
josuah, line 3: Is this not to be replaced by memset so we can remove this file entirely?

```
#define byte_zero(s, n) memset(s, 0, n)
```

byte.h
```diff
@@ -3,11 +3,11 @@
...
-extern unsigned int byte_chr();
-extern unsigned int byte_rchr();
-extern void byte_copy();
-extern void byte_copyr();
-extern void byte_zero();
+extern unsigned int byte_chr(char *s, unsigned int n, int c);
+extern unsigned int byte_rchr(char *s, unsigned int n, int c);
+extern void byte_copy(char *to, unsigned int n, char *from);
+extern void byte_copyr(char *to, unsigned int n, char *from);
+extern void byte_zero(char *s, unsigned int n);
```
josuah, line 11: http://git.skarnet.org/cgi-bin/cgit.cgi/skalibs/tree/src/include/skalibs/bytestr.h

This above provides example on how to convert DJB-style functions to <string.h> ones.

byte_zero.c
DerDakon: I intentionally kept this so the compiler will not optimize the zeroing aways. I need a lot of logic to get this portable in Qsmtp, and there is still the possibility that it will fallback to not finding one: https://github.com/DerDakon/Qsmtp/blob/master/qsmtpd/auth.c#L29

byte.h
DerDakon: Not in this PR ;)

byte_zero.c
josuah: Would a comment be welcome on the source so that we state out where we want *not* to edit it?  Although maybe `byte_zero` is a famous enough topic.

https://github.com/libressl-portable/openbsd/issues/5

It looks like there is a hard time getting this not optimized away.

DerDakon: Our luck is that we don't use -flto at the moment. And the other thing is that it is used in that "dangerous" way only in qmail-popup, which will hopefully just go away together with qmail-pop3d.

josuah:
Same logics would go for the case_*() functions.

Makefile
```diff
@@ -1718,6 +1718,13 @@ byte_cr.o byte_zero.o
 	str_chr.o str_rchr.o str_start.o byte_chr.o byte_rchr.o \
 	byte_copy.o byte_cr.o byte_zero.o
 
+str_chr.c: \
+chkstrchr0.c compile load str_chr.c.impl str_chr.c.gnu
+	 ( ( ./compile chkstrchr0.c  && ./load chkstrchr0 && \
+	./chkstrchr0 ) >/dev/null 2>&1 \
+	&& cat str_chr.c.gnu || cat str_chr.c.impl ) > str_chr.c
+	rm -f chkstrchr0.o chkstrchr0
+
```
schmonz, line 1723: My vote is that this commit is a net increase in complexity for unclear value, and we should stick with the simple-enough, fast-enough, and universally available `strchr()` + `strlen()` implementation.

str.h
```diff
@@ -9,7 +9,10 @@
 #define str_len(s) strlen((s))
 extern unsigned int str_chr();
 extern unsigned int str_rchr();
-extern int str_start();
+static inline int str_start(const char *s, const char *t)
+{
+  return strncmp(s, t, strlen(t)) == 0;
```
schmonz, line 22: I don't know which systems have `strncmp()` and which don't. How universally available is it?

schmonz: Possibly better as a `#define`?

DerDakon: My manpage says:

```
CONFORMING TO
       POSIX.1-2001, POSIX.1-2008, C89, C99, SVr4, 4.3BSD.
```

DerDakon: This is intentionally no define because with a inline function is is easy to be able to use "t" twice, with a macro one would have to add additional tricks to make it save that the expression isn't evaluated twice, for no real benefit.

qmail-remote.c
```diff
@@ -287,20 +288,20 @@ stralloc canonbox = {0};
 
 void addrmangle(stralloc *saout, char *s)
 {
-  int j;
+  char *at;
  
-  j = str_rchr(s,'@');
-  if (!s[j]) {
+  at = strrchr(s,'@');
+  if (!at) {
```
schmonz, line 294: The way you're testing the `str_*` interfaces before changing their implementations makes me very comfortable with those changes. These changes make me nervous without tests around callers. I'm further worried about breaking patches sooner than we have to, particularly for `qmail-remote`. My vote is to save the idea in this commit for later.

tests/unittest_case.c
```diff
@@ -0,0 +1,48 @@
+#include <check.h>
+
+#include "case.h"
+
+START_TEST(test_case_diffs)
+{
+  ck_assert_uint_eq(case_diffs("abc", "abc"), 0);
+  ck_assert_uint_ne(case_diffs("aba", "ab"), 0);
```
schmonz, line 8: Suggest adding `ck_assert_uint_ne(case_diffs("ab", "aba"), 0);`

tests/unittest_case.c
```diff
@@ -0,0 +1,48 @@
...
+  ck_assert_uint_eq(case_diffs("abc", "AbC"), 0);
+  ck_assert_uint_eq(case_diffs("abc", "ABC"), 0);
+  ck_assert_uint_eq(case_diffs("Abc", "abC"), 0);
+  ck_assert_uint_ne(case_diffs("ä", "Ä"), 0);
+  ck_assert_uint_ne(case_diffs("", ""), 1);
```
schmonz, line 13: Another example or two, please, of strings that `case_diffs()` would say 1 about

case.h
```diff
@@ -1,9 +1,12 @@
 #ifndef CASE_H
 #define CASE_H
 
+#include <string.h>
+#include <strings.h>
+
 extern void case_lowers();
 extern void case_lowerb();
-extern int case_diffs();
+#define case_diffs(s,t) (strcasecmp((s),(t)) != 0)
```
schmonz, line 9: Why `strncmp()` elsewhere but not `strncasecmp()` here, or vice versa? Serious question. When do we trust non-`n` functions, when would we prefer to call `strlen()` and pass that in, and when would we prefer to already have the length and pass _that_ in?

schmonz (changes requested):
I'd like to leave out a couple commits (hopefully review shows clearly enough which ones) and merge the rest pronto. Adding tests that let us safely delete needlessly custom code might be my favorite kind of change.

case.h
DerDakon: because ```strncasecmp()``` is nonstandard and may not be available everywhere.

schmonz: Is `strcasecmp` not historically one of the sometimes dangerous libc string functions? I don’t know which those were and I don’t know how to find out, either.

qmail-remote.c
DerDakon: I have put that commit in it's own branch.

schmonz:
Is this ready for re-review?

case_lowerb.c
DerDakon: Yes, I would like to see it in a different PR on top of this, best once this is merged to avoid duplicate work.

DerDakon:
I think so. One obvious lesson from the last PRs is that ```case_startb()``` should not be deleted, but only deprecated. I'll include that once I rebase this branch, but otherwise it should be pretty fine.

DerDakon:
```case_startb()``` never existed, only the declaration. So I just wiped it from the man page, too.

File: byte_zero.c (hunk 4 above)
Comment (schmonz) [reply byte_zero.c:3]: Cross-referencing #65 which at one point was suggesting

```c
#define byte_zero(s, n) memset((s), 0, (n))
```

now that I see this conversation here :-)

(Hopefully not derailing this PR: I know in our last release we expressed intent to delete `qmail-popup` and `qmail-pop3d`, but I'm not ready to do that yet, even though I don't run them myself. They're a worked example of a particularly qmail-ish design we might want to follow for AUTH and extend to support TLS, so I'd like us to hang onto them at least until we've added those features -- assuming they're relatively cheap to carry with us a bit longer, which I think is true.)

case.h
DerDakon: Those like ```strcpy()``` where there is possibility of overrun or ```strncpy()``` where the terminating 0-byte may not be written.

DerDakon:
ping

DerDakon:
I now also took care of ```byte_copy()``` and ```byte_copyr()```.

mbhangui (approved):

josuah (approved):
Thankfully DJB introduced functions compatible with the libc, so this gets straightforward: a shallow renaming.

PR #144, opened 2020-05-24
make most output of auto-str readable

DerDakon:
Avoid recoding letters and anything from "/.-_" into octal, just print them directly.

### DIFF
--- a/auto-str.c
+++ b/auto-str.c
@@ -1,3 +1,4 @@
+#include <ctype.h>
 #include "substdio.h"
 #include "readwrite.h"
 #include "exit.h"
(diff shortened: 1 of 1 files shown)


auto-str.c
```diff
@@ -30,6 +30,11 @@ char **argv;
   puts("[] = \"\\\n");
 
   while (ch = *value++) {
+    if ((ch >= 'a' && ch <= 'z') || (ch >= 'A' && ch <= 'Z') || ch == '/' || ch == '_' || ch == '-' || ch == '.') {
```
schmonz, line 33: I like where this is going. How about extracting `is_legible()` and instead of `continue`, `if (is_legible(ch)) { //print verbatim } else { //octal it up }` ?

leahneukirchen: use isalnum?

DerDakon: ```is_alpha()``` maybe. I decided against numbers because it becomes hard to read a number after an octal literal. I used continue so I don't have to change the other lines, but I can do that of course.

DerDakon: According to the documentation ```isalpha()``` is locale dependent, which is something I would like to avoid.

leahneukirchen: We don't use setlocale, so this is fine for the C locale.

DerDakon: I'm unsure what happens if that program is run in an environment which has a locale set.

leahneukirchen: > On startup of the main program, the portable "C" locale is selected as default.

PR #145, opened 2020-05-24
return int from all main() functions

DerDakon:


bouncesaying.c
```diff
@@ -7,14 +7,12 @@
...
-void main(argc,argv)
-int argc;
-char **argv;
+int main(int argc, char **argv)
 {
   int pid;
   int wstat;
 
-  if (!argv[1])
+  if (argc == 1 || !argv[1])
```
josuah, line 15: I did not know there could be any situation `!argv[1] && argc ==1` could happen.

forward.c
```diff
@@ -25,13 +25,12 @@ substdio ssout = SUBSTDIO_FDBUF(mywrite,-1,outbuf,sizeof outbuf);
...
 
-void main(argc,argv)
-int argc;
-char **argv;
+int main(int argc, char **argv)
 {
   char *sender;
   char *dtline;
   char *qqx;
+  int i;
```
josuah, line 33: Given `i` is also declared in `for (int i; ...; ...) { ... }`, is it really needed here?
Sorry if I missed it.

hostname.c
```diff
@@ -1,17 +1,16 @@
 #include "substdio.h"
 #include "subfd.h"
 #include "readwrite.h"
-#include "exit.h"
 
 char host[256];
 
-void main()
+int main()
```
josuah, line 7: while at it, do we also convert it to `int main(void)` ?

bouncesaying.c
leahneukirchen: I think just checking argc is good enough?

DerDakon: It probably is. Let me do some experiments.

chkspawn.c
```diff
@@ -2,7 +2,6 @@
 #include "subfd.h"
 #include "fmt.h"
 #include "select.h"
-#include "exit.h"
```
schmonz, line 5: `Makefile` still has `chkspawn.o` depending on `exit.h`

dnsfq.c
```diff
@@ -5,26 +5,25 @@
 #include "dnsdoe.h"
 #include "ip.h"
 #include "ipalloc.h"
-#include "exit.h"
```
schmonz, line 8: Makefile dependency thing again

dnsip.c
```diff
@@ -5,7 +5,6 @@
 #include "dnsdoe.h"
 #include "ip.h"
 #include "ipalloc.h"
-#include "exit.h"
```
schmonz, line 8: You already got the idea last time :-)

auto-int.c
```diff
@@ -13,15 +13,14 @@ char *s;
...
-int argc;
-char **argv;
+int main(int argc, char **argv)
 {
   char *name;
   char *value;
   unsigned long num;
   char strnum[FMT_ULONG];
 
+  if (argc != 3) _exit(100);
```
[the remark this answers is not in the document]
DerDakon: -Wunused-parameter or so

PR #146, opened 2020-06-06
fix all warnings about usage of undeclared functions

DerDakon:


seek.h
```diff
@@ -1,14 +1,35 @@
...
+
+
+static inline int seek_end(int fd)
+{
+  if (lseek(fd, 0, SEEK_END) == -1)
+    return -1;
+  return 0;
+}
+
+static inline int seek_trunc(int fd, seek_pos pos)
```
schmonz, line 29: Out of curiosity, why not `#define seek_trunc(fd, pos) (ftruncate(fd, pos))`, at least for this one with no branching logic?

maildir2mbox.c
```diff
@@ -14,6 +15,8 @@
 #include "myctime.h"
 #include "maildir.h"
 
+extern int rename(const char *, const char *);
```
schmonz, line 18: Are you already plotting the demise of `puts()`, so we can come back for `<stdio.h>` on these?

DerDakon: I would love to, but I think ```puts()``` is one of the things that is likely actually used by external patches.

seek.h
DerDakon: sure, would be consistent with other changes.

maildir2mbox.c
schmonz: Yes, that’s true.

I’m not worried we’ll forget to come back to this someday (when external patches are way less common). Pretty soon, when we have hardly any `extern` left, this’ll stick out. :-)

PR #148, opened 2020-06-11
Symlink "CI" configs somewhere easier to see.

schmonz:
I don't think about our autobuilders as often as I should. When I do, I'm never sure I'm looking at all of them. This might help me.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 .cirrus.yml export-ignore
 .git* export-ignore
 .travis.yml export-ignore
+autobuilds export-ignore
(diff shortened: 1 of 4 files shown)


DerDakon (changes requested):
Please also add it to .gitattributes as there is no value in having that in the release tarball when we also filter out the actual configs.

Also I'm thinking about naming it maybe "autobuilds" to make it a bit more obvious. Opinions?

schmonz:
> Please also add it to .gitattributes as there is no value in having that in the release tarball when we also filter out the actual configs.
> 
> Also I'm thinking about naming it maybe "autobuilds" to make it a bit more obvious. Opinions?

Agree. I've done both.

DerDakon (approved):

mbhangui (approved):

PR #149, opened 2020-06-11
remove duplication in Makefile

DerDakon:
Basically the same what we have already done for catman targets.


Makefile
```diff
@@ -109,81 +109,47 @@ auto_usera.c: \
...
-	./compile auto_userr.c
-
 auto_users.c: \
 auto-str conf-users
 	./auto-str auto_users `head -n 8 conf-users | tail -n -1` > auto_users.c
 
-auto_users.o: \
-compile auto_users.c
-	./compile auto_users.c
+auto_user%.o: auto_user%.c compile
```
schmonz, line 140: I’d preserve the ordering `auto_user%.o: compile auto_user%.c`

Makefile
```diff
@@ -205,84 +205,34 @@ bouncesaying.o: \
...
-	./compile byte_cr.c
-
-byte_rchr.o: \
-compile byte_rchr.c byte.h
-	./compile byte_rchr.c
-
-byte_zero.o: \
-compile byte_zero.c byte.h
-	./compile byte_zero.c
+byte_%.o: byte_%.c compile byte.h
```
schmonz, line 208: Also prefer keeping `compile` first in the dep list

Makefile
DerDakon: Then $< would refer to compile and it would call ```./compile compile```.

schmonz: Achaaaa. Okay yep

josuah:
I thought the % makefiles rules were not portable because they are not POSIX, but they are at least working on BSD make : https://notqmail.z0.is/patch/0-patch/85555302dd1f6688babb236eb364e9e6f08c96ac/index.make.out

schmonz:
My interpretation was the opposite. Interesting. I was thinking anything that highlights patterns in our existing dependencies is a step toward automation. But I’m not actually using redo, and you are. Does this change feel counterproductive to you, or not obviously worth the cost, or something like that?

DerDakon:
Could someone please test if this actually _honors_ both rules? I had the impression that BSD make was only looking at the compile-rule, but ignoring the later dependency rules.

Makefile
```diff
@@ -21,6 +21,9 @@ default: it
 .8.0:
 	$(NROFF) -man $< >$@
 
+%.o: %.c compile
```
josuah, line 24: A slightly less readable but more portable way would be `.c.o:`, but I doubt this would support adding `compile` as dependency of `.o`.

josuah (approved):
Note that this removes dependency on headers: if they get modified and the project is recompiled, the associated .o might not be recompiled.  Did nowaday NVMe disks made make clean all cheap enough?

This also introduces `%` rules that are not POSIX, but there are already common extensions used.

DerDakon:
> Note that this removes dependency on headers: if they get modified and the project is recompiled, the associated .o might not be recompiled. Did nowaday NVMe disks made make clean all cheap enough?

This was basically my question. This is not intended, and it works with GNU make, but if BSD make doesn't support this I can scrap half of it because that is a regression.

josuah:
https://notqmail.z0.is/commit/03b8497487b7ebc0580d92059ae77171fff71601/patch/0/log

Works without trouble on that instance.

mbhangui (approved):

PR #151, opened 2020-06-13
replace alloc_re() with plain realloc()

DerDakon:
It is only used at 2 places, and needs overhaul anyway to get rid of compiler warnings. See more rationale in the commit messages.


DerDakon:
I assume that he want a simple interface where he could check the return value and does not have to assign anything afterwards, which is indeed a bit cumbersome for realloc().

PR #152, opened 2020-06-14
use NULL instead of casting 0 to a pointer type

DerDakon:
The conversion was done using:

```shell
sed -r 's/\([^()]* \*\)[[:space:]]*0([,\)])/NULL\1/g' -i *.[ch]
```


josuah:
Thank you for pointing it out, I learned that today. DJB expressedly avoided variadic arguments.

PR #153, opened 2020-06-15
add additional groups for qmail-remote

DerDakon:
The idea is that qmail-remote may read files that other qmail processes may not read, and additionally ones that it may exclusively read without owning them. The intended use case is for DKIM keys (exclusively read) and TLS keys (shared with qmail-smtpd).

### DIFF
--- a/prot.c
+++ b/prot.c
@@ -1,12 +1,21 @@
 #include "prot.h"
 
 #include <sys/types.h>
-#include <unistd.h>
 #include <grp.h>
+#include <unistd.h>
 
 int prot_gid(gid_t gid)
 {
   if (setgroups(1,&gid) == -1) return -1;
+
+  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
+}
+
+int prot_gids(const char *user, gid_t gid)
+{
+  /* member of too many groups */
+  if (initgroups(user, gid) == -1) return -1;
+
   return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
 }
 
--- a/prot.c
+++ b/prot.c
@@ -18,8 +18,3 @@ int prot_gids(const char *user, gid_t gid)
 
   return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
 }
-
-int prot_uid(uid_t uid)
-{
-  return setuid(uid);
-}
--- a/qmail-start.c
+++ b/qmail-start.c
@@ -105,6 +105,7 @@ int main(int argc, char **argv)
   switch(fork()) {
     case -1: die();
     case 0:
+      if (prot_gids(auto_userr, auto_gidq) == -1) die();
       if (prot_uid(auto_uidr) == -1) die();
       if (fd_copy(0,pi3[0]) == -1) die();
       if (fd_copy(1,pi4[1]) == -1) die();
--- a/.github/workflows/openbsd.yml
+++ b/.github/workflows/openbsd.yml
@@ -1,0 +1,49 @@
+name: C CI on OpenBSD
+
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: OpenBSD-${{ matrix.config.name }}
+    # see https://github.com/vmactions/openbsd-vm: must be macos-12
+    runs-on: macos-12
+    env:
+      CFLAGS: ${{ matrix.config.cflags }}
+      NROFF: ${{ matrix.config.nroff }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+         - {
+           name: "default64",
+           nroff: "nroff",
+           cflags: ""
+         }
+         - {
+           name: "no_obsolete64",
+           nroff: "true",
+           cflags: "-DDEPRECATED_FUNCTIONS_REMOVED"
+         }
+
+    steps:
+    - name: checkout
+      uses: actions/checkout@v3
+
+    - name: Test in OpenBSD
+      id: test
+      uses: vmactions/openbsd-vm@v0
+      with:
+        envs: 'CFLAGS NROFF'
+        prepare: |
+          pkg_add check
+        run: |
+          echo "cc -m64 -O2 ${CFLAGS}" > conf-cc
+          echo "cc -m64 -s" > conf-ld
+          make it man NROFF=${NROFF}
+          cd tests
+          make test
--- a/.github/workflows/solaris.yml
+++ b/.github/workflows/solaris.yml
@@ -10,7 +10,8 @@ on:
 jobs:
   build:
     name: Solaris-${{ matrix.config.name }}
-    runs-on: macos-10.15
+    # see https://github.com/vmactions/solaris-vm: must be macos-12
+    runs-on: macos-12
     env:
       CFLAGS: ${{ matrix.config.cflags }}
       NROFF: ${{ matrix.config.nroff }}
@@ -32,9 +33,9 @@ jobs:
 
     steps:
     - name: checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
-    - name: Test in solaris
+    - name: Test in Solaris
       id: test
       uses: vmactions/solaris-vm@v0
       with:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,4 @@
+- 20221115 honor all groups of qmailr user when spawning qmail-rspawn
 - 20221019 code: netstring parsing in qmail-qmtpd & qmail-qmqpd now
            checks that the netstring length is actually a number.
 - 20210122 code: remove register storage class declaration from codebase.
--- a/autobuilds/openbsd.yml
+++ b/autobuilds/openbsd.yml
@@ -1,0 +1,1 @@
+../.github/workflows/openbsd.yml
--- a/prot.h
+++ b/prot.h
@@ -2,9 +2,10 @@
 #define PROT_H
 
 #include <sys/types.h>
+#include <unistd.h>
 
 extern int prot_gid(gid_t gid);
 extern int prot_gids(const char *user, gid_t gid);
-extern int prot_uid(uid_t uid);
+#define prot_uid(uid) setuid(uid)
 
 #endif
--- a/prot.h
+++ b/prot.h
@@ -1,7 +1,10 @@
 #ifndef PROT_H
 #define PROT_H
 
-extern int prot_gid();
-extern int prot_uid();
+#include <sys/types.h>
+
+extern int prot_gid(gid_t gid);
+extern int prot_gids(const char *user, gid_t gid);
+extern int prot_uid(uid_t uid);
 
 #endif


mbhangui:
Why not use /etc/group to gain additional privileges. read groups to which qmailr belongs, using getgrent()? We can write a function get calls getgrent(), get all supplementary groups and set it using setgroups(). Using /etc/groups would make the solution flexible. 

```
/*-
 * scan the group file for all supplementary groups.
 * Return NULL if unsucessfull
 * The returned value should be freed after use
 */
gid_t   *
grpscan(char *user, int *ngroups)
{   
    struct passwd  *pwd;
    struct group   *grp;
    long            maxgroups, idx;
    gid_t          *gidsetlen;
    char          **ptr;

    if (!user || !*user)
        return ((gid_t *) 0);
    if ((maxgroups = sysconf(_SC_NGROUPS_MAX)) == -1)
        return ((gid_t *) 0);
    else
    if (!(gidsetlen = (gid_t *) alloc(maxgroups * sizeof(gid_t))))
        return ((gid_t *) 0);
    else
    if (!(pwd = getpwnam(user)))
        return ((gid_t *) 0);
    idx = 0;
    gidsetlen[idx++] = pwd->pw_gid; /* the base gid */
    endpwent();
    for (;;) {
        if (!(grp = getgrent()))
            break;
    struct group   *grp;
    long            maxgroups, idx;
    gid_t          *gidsetlen;
    char          **ptr;

    if (!user || !*user)
        return ((gid_t *) 0);
    if ((maxgroups = sysconf(_SC_NGROUPS_MAX)) == -1)
        return ((gid_t *) 0);
    else
    if (!(gidsetlen = (gid_t *) alloc(maxgroups * sizeof(gid_t))))
        return ((gid_t *) 0);
    else
    if (!(pwd = getpwnam(user)))
        return ((gid_t *) 0);
    idx = 0;
    gidsetlen[idx++] = pwd->pw_gid; /* the base gid */
    endpwent();
    for (;;) {
        if (!(grp = getgrent()))
            break;
        for (ptr = grp->gr_mem; *ptr; ptr++) {
            if (!str_diff(user, *ptr) && grp->gr_gid != gidsetlen[0])
                gidsetlen[idx++] = grp->gr_gid; /* supplementary group ids */
        }
    }
    endgrent();
    *ngroups = idx;
    return (gidsetlen);
}
```
in qmail-start, we can do
```
gid_t   *gidset;
int         ngroups;

if (!(gidset = grpscan("qmailr", &ngroups))
  die();
if (setgroups(ngroups, gidset))
   die();
```

DerDakon:
I think I'll simply use ```getgrouplist()```.

mbhangui:
> I think I'll simply use `getgrouplist()`.

Yes. SImpler. Never knew that this function existed :)

mbhangui:
I needed something like this for few things apart from qmail-remote. In qmail-smtpd. So the logical place to do the same thing would be in setuidgid.c to use /etc/groups, to set all supplementary groups for the account.

DerDakon:
qmail-smtpd is not started from qmail-start.

mbhangui:
> qmail-smtpd is not started from qmail-start.

Yes, I should have been clear about that. I use setuidgid (from daemontools) in run file to start qmail-smtpd. So I am making similar changes in setuidgid.c to achieve this.

DerDakon:
I added a last commit that uses an object file for all the user and group name constants. That one is optional, either we take this one and I squash it into the first, or we drop it and take it only the first one. Or not even this, depends on you.

prot.c
```diff
@@ -9,7 +9,19 @@ int prot_gid(gid_t gid)
...
-  return setuid(uid);
+  gid_t gids[GROUP_COUNT];
+  int gcount = GROUP_COUNT;
+  int r = getgrouplist(user, gid, gids, &gcount);
+  /* member of too many groups */
+  if (r < 0)
+    return r;
+  if (setgroups(gcount,gids) == -1) return -1;
+  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
 }
```
schmonz, line 27: Untested, but how about this attempted refactoring of `prot.h` and `prot.c`:

```c
#ifndef PROT_H
#define PROT_H

#include <sys/types.h>
#include <unistd.h>

#define prot_gid(gid) prot_gids(NULL, gid);
extern int prot_gids(const char *user, gid_t gid);
#define prot_uid(uid) setuid(uid)

#endif
```

```c
#include <sys/types.h>
#include <unistd.h>
#include <grp.h>
#include "prot.h"

#ifndef GROUP_COUNT
/* use a sane default, but one can override it from outside if there are
 * higher requirements. */
#define GROUP_COUNT 8
#endif
int prot_gids(const char *user, gid_t gid)
{
  gid_t gids[GROUP_COUNT];
  int gcount = GROUP_COUNT;

  if (user) {
    int r = getgrouplist(user, gid, gids, &gcount);
    /* member of too many groups */
    if (r < 0)
      return r;
  } else {
    gcount = 1;
    gids[0] = gid;
  }

  if (setgroups(gcount,gids) == -1) return -1;
  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
}
```

schmonz (commented):
Intuitively, I like how many TARGETS go away with `uids.o`, but it seems a little cleverer than `uids.a`. Either way, we're obscuring one way for code-readers can know exactly which IDs are used where. But there are other ways to find out. I don't think I have a strong feeling in any direction (keep as is, switch to .a, switch to .o).

qmail-start.c
```diff
@@ -103,6 +103,7 @@ int main(int argc, char **argv)
   switch(fork()) {
     case -1: die();
     case 0:
+      if (prot_gids(auto_userr, auto_gidq) == -1) die();
```
schmonz, line 106: Do we want to do the equivalent for other spawned programs as well?

prot.c
DerDakon: Lovely.

qmail-start.c
DerDakon: It is already there: the prot_gid() call right after the auto_gid() calls.

Makefile
```diff
@@ -1386,11 +1298,11 @@ error.h
...
 load qmail-rspawn.o spawn.o tcpto_clean.o coe.o sig.a open.a \
 lock.a wait.a fd.a stralloc.a substdio.a error.a env.a str.a \
-auto_qmail.o uid.o auto_userq.o auto_spawn.o
+auto_qmail.o uid.o auto_spawn.o uids.o
 	./load qmail-rspawn spawn.o tcpto_clean.o coe.o \
 	sig.a open.a lock.a wait.a fd.a stralloc.a \
-	auto_qmail.o uid.o auto_userq.o substdio.a error.a env.a str.a \
-	auto_spawn.o
+	auto_qmail.o uid.o substdio.a error.a env.a str.a \
+	auto_spawn.o uids.o
```
schmonz, line 1305: Reordered

Makefile
```diff
@@ -1439,13 +1351,10 @@ qmail-send.service.in conf-qmail
...
-auto_usero.o auto_userp.o auto_userq.o auto_userr.o auto_users.o auto_groupn.o \
-auto_groupq.o control.o open.a getln.a stralloc.a substdio.a error.a \
-str.a fs.a auto_qmail.o auto_break.o auto_patrn.o auto_spawn.o auto_split.o
-	./load qmail-showctl uid.o gid.o auto_usera.o auto_userd.o \
-	auto_userl.o auto_usero.o auto_userp.o auto_userq.o auto_userr.o \
-	auto_users.o auto_groupn.o auto_groupq.o control.o open.a getln.a \
+load qmail-showctl.o uid.o gid.o control.o open.a getln.a stralloc.a \
+substdio.a error.a uids.o str.a fs.a auto_qmail.o auto_break.o auto_patrn.o \
+auto_spawn.o auto_split.o
+	./load qmail-showctl uid.o gid.o control.o open.a getln.a uids.o \
```
schmonz, line 1357: Both dependency and command slightly reordered

Makefile
```diff
@@ -1484,11 +1393,8 @@ exit.h rcpthosts.h timeoutread.h timeoutwrite.h commands.h
 	./compile qmail-smtpd.c
 
 qmail-start: \
-load qmail-start.o prot.o fd.a uid.o gid.o auto_userl.o auto_userq.o \
-auto_userr.o auto_users.o auto_groupn.o auto_groupq.o substdio.a error.a str.a
-	./load qmail-start prot.o fd.a uid.o gid.o auto_userl.o auto_userq.o \
-	auto_userr.o auto_users.o auto_groupn.o auto_groupq.o substdio.a \
-	error.a str.a
+load qmail-start.o prot.o fd.a uid.o gid.o substdio.a error.a str.a uids.o
+	./load qmail-start prot.o fd.a uid.o gid.o substdio.a error.a str.a uids.o
```
schmonz, line 1397: Same

Makefile
```diff
@@ -1923,6 +1829,24 @@ uid.o: \
 compile uid.c uidgid.h subfd.h substdio.h exit.h
 	./compile uid.c
 
+uids.c: \
+auto-str conf-users
+	./auto-str auto_usera `head -n 1 conf-users` > uids.c.tmp
```
schmonz, line 1834: Possibly goofy idea, but for visual consistency, maybe pipe this (and `auto_groupq`) through `tail -n -1` also?

TARGETS
```diff
@@ -141,26 +143,6 @@ qmail-lspawn
...
-auto_userq.c
-auto_userq.o
-auto_userr.c
-auto_userr.o
-auto_users.c
-auto_users.o
-auto_groupn.c
-auto_groupn.o
-auto_groupq.c
-auto_groupq.o
```
schmonz, line 163: This is great

prot.c
```diff
@@ -1,15 +1,28 @@
...
+/* use a sane default, but one can override it from outside if there are
+ * higher requirements. */
+#define GROUP_COUNT 8
+#endif
+int prot_gids(const char *user, gid_t gid)
 {
-  if (setgroups(1,&gid) == -1) return -1;
-  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
-}
+  gid_t gids[GROUP_COUNT];
```
schmonz, line 13: I guess this is a little space-wasteful in the `prot_gid()` case, but meh

schmonz (changes requested):
If we'd been reviewing PRs more actively, I might want the `ids.o` bit as its own PR. But we haven't been and I don't mind it going in as part of this change.

Makefile
```diff
@@ -711,24 +631,19 @@ install:
...
 instcheck: \
 load instcheck.o instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o uid.o \
-gid.o auto_usera.o auto_usero.o auto_userp.o auto_userq.o auto_userr.o \
-auto_users.o auto_groupq.o strerr.a substdio.a error.a str.a fs.a
+gid.o uids.o strerr.a substdio.a error.a str.a fs.a
 	./load instcheck instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o \
-	uid.o gid.o auto_usera.o auto_usero.o auto_userp.o auto_userq.o \
-	auto_userr.o auto_users.o auto_groupq.o strerr.a substdio.a error.a \
-	str.a fs.a
+	uid.o gid.o uids.o strerr.a substdio.a error.a str.a fs.a
```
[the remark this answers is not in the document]
DerDakon: Sounds sane. I don't mind splitting it into it's own PR.

Makefile
```diff
@@ -1132,12 +1047,10 @@ qmail-log.5
 qmail-lspawn: \
 load qmail-lspawn.o spawn.o prot.o slurpclose.o coe.o sig.a wait.a \
 case.a cdb.a fd.a open.a stralloc.a substdio.a error.a str.a \
-fs.a auto_qmail.o uid.o gid.o auto_userp.o auto_userq.o auto_groupn.o \
-auto_spawn.o
+fs.a uids.o auto_qmail.o uid.o gid.o
```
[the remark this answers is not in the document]
DerDakon: Will fix that.

prot.c
DerDakon: For a 32 bit gid_t these are 32 byte of stack space - costs us basically nothing. There are worse things that happen on the stack and it is freed upon exit of this function, so who cares?

Makefile
DerDakon: I have added some spaces for alignment.

Makefile
```diff
@@ -1128,12 +1123,10 @@ qmail-log.5
 qmail-lspawn: \
 load qmail-lspawn.o spawn.o prot.o slurpclose.o coe.o sig.a wait.a \
 case.a cdb.a fd.a open.a stralloc.a substdio.a error.a str.a \
-fs.a auto_qmail.o uid.o gid.o auto_userp.o auto_userq.o auto_groupn.o \
-auto_spawn.o
+fs.a ids.a auto_qmail.o uid.o gid.o
```
schmonz: Nitpick: `ids.a` after `gid.o`

Makefile
```diff
@@ -1128,12 +1123,10 @@ qmail-log.5
...
 	./load qmail-lspawn spawn.o prot.o slurpclose.o coe.o sig.a wait.a \
 	case.a cdb.a fd.a open.a stralloc.a auto_qmail.o uid.o gid.o \
-	auto_userp.o auto_userq.o auto_groupn.o auto_spawn.o substdio.a \
-	error.a str.a fs.a
+	auto_spawn.o substdio.a ids.a error.a str.a fs.a
```
schmonz: Nitpick: `ids.a` before `auto_spawn.o`

Makefile
```diff
@@ -1382,10 +1374,10 @@ error.h
 qmail-rspawn: \
 load qmail-rspawn.o spawn.o tcpto_clean.o coe.o sig.a open.a \
 lock.a wait.a fd.a stralloc.a substdio.a error.a env.a str.a \
-auto_qmail.o uid.o auto_userq.o auto_spawn.o
+auto_qmail.o uid.o auto_spawn.o ids.a
```
schmonz: Nitpick: `ids.a` before `auto_spawn.o`

Makefile
```diff
@@ -1435,13 +1427,10 @@ qmail-send.service.in conf-qmail
...
 
 qmail-showctl: \
-load qmail-showctl.o uid.o gid.o auto_usera.o auto_userd.o auto_userl.o \
-auto_usero.o auto_userp.o auto_userq.o auto_userr.o auto_users.o auto_groupn.o \
-auto_groupq.o control.o open.a getln.a stralloc.a substdio.a error.a \
-str.a fs.a auto_qmail.o auto_break.o auto_patrn.o auto_spawn.o auto_split.o
-	./load qmail-showctl uid.o gid.o auto_usera.o auto_userd.o \
-	auto_userl.o auto_usero.o auto_userp.o auto_userq.o auto_userr.o \
-	auto_users.o auto_groupn.o auto_groupq.o control.o open.a getln.a \
+load qmail-showctl.o uid.o gid.o control.o open.a getln.a stralloc.a \
```
schmonz: Nitpick: `ids.a` before `control.o`

Makefile
```diff
@@ -1435,13 +1427,10 @@ qmail-send.service.in conf-qmail
...
+substdio.a error.a ids.a str.a fs.a auto_qmail.o auto_break.o auto_patrn.o \
+auto_spawn.o auto_split.o
+	./load qmail-showctl uid.o gid.o control.o open.a getln.a ids.a \
```
schmonz: Same nitpick as previous

Makefile
```diff
@@ -1480,11 +1469,8 @@ exit.h rcpthosts.h timeoutread.h timeoutwrite.h commands.h
 	./compile qmail-smtpd.c
 
 qmail-start: \
-load qmail-start.o prot.o fd.a uid.o gid.o auto_userl.o auto_userq.o \
-auto_userr.o auto_users.o auto_groupn.o auto_groupq.o substdio.a error.a str.a
-	./load qmail-start prot.o fd.a uid.o gid.o auto_userl.o auto_userq.o \
-	auto_userr.o auto_users.o auto_groupn.o auto_groupq.o substdio.a \
-	error.a str.a
+load qmail-start.o prot.o fd.a uid.o gid.o ids.a substdio.a error.a str.a
+	./load qmail-start prot.o fd.a uid.o gid.o ids.a substdio.a error.a str.a
```
schmonz: Elegant

Makefile
```diff
@@ -1919,6 +1905,13 @@ uid.o: \
 compile uid.c uidgid.h subfd.h substdio.h exit.h
 	./compile uid.c
 
+ids.a: \
+makelib auto_usera.o auto_userd.o auto_userl.o auto_usero.o auto_userp.o \
+auto_userq.o auto_userr.o auto_users.o auto_groupn.o auto_groupq.o
+	./makelib ids.a auto_usera.o auto_userd.o auto_userl.o auto_usero.o \
+	auto_userp.o auto_userq.o auto_userr.o auto_users.o auto_groupn.o \
+	auto_groupq.o
```
schmonz: Yep, that's all of 'em

prot.c
```diff
@@ -1,11 +1,29 @@
+#include "prot.h"
```
schmonz, line 1: Better after the `<>` includes?

schmonz (changes requested):
Were we going to do `ids.o` as its own PR? It's a good one on its own, and then this PR gets nice and short.

Makefile
DerDakon: I tried to keep groups of archives or object files together whereever possible, even if that means I shift these things around a bit. If we go for the .o solution I would squash the last commit into the first one and entirely get rid of the intermediate state, so this will not persist.

prot.c
DerDakon: The idea is that this enforces prot.h to be self-contained. It will fail to build if it misses a type or something similar. If you put it after the system headers they could provide the missing types and hide the error.

DerDakon:
I moved all the Makefile fiddling into #184.

prot.c
schmonz: Makes sense. Let's try to do this in more places.

schmonz (approved):
This one should get a CHANGES entry.

prot.c
```diff
@@ -1,11 +1,29 @@
...
+int prot_gids(const char *user, gid_t gid)
 {
-  if (setgroups(1,&gid) == -1) return -1;
+  gid_t gids[GROUP_COUNT];
+  int gcount = GROUP_COUNT;
+
+  if (user) {
+    int r = getgrouplist(user, gid, gids, &gcount);
+    /* member of too many groups */
+    if (r < 0)
```
schmonz, line 19: Given the nested ifs, I'd slightly prefer braces around this one.

schmonz (approved):

prot.c
```diff
@@ -1,15 +1,29 @@
...
-  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
-}
+  gid_t gids[GROUP_COUNT];
+  int gcount = GROUP_COUNT;
 
-int prot_uid(uid_t uid)
-{
-  return setuid(uid);
+  if (user) {
+    int r = getgrouplist(user, gid, gids, &gcount);
```
leahneukirchen, line 17: We could just use `initgroups` no?

prot.c
```diff
@@ -1,15 +1,29 @@
...
+    if (r < 0) {
+      return r;
+    }
+  } else {
+    gcount = 1;
+    gids[0] = gid;
+  }
+
+  if (setgroups(gcount,gids) == -1) return -1;
+  return setgid(gid); /* _should_ be redundant, but on some systems it isn't */
```
leahneukirchen, line 28: We should always setgid, as setgroups only deals with secondary groups (this also makes the else-clause above useless).

leahneukirchen (changes requested):

prot.c
leahneukirchen: Please verify getgrouplist works on MacOS, as getgrouplist uses int* in the prototype there. But sizeof gid_t is 4, so it should work. I still recommend to use initgroups tho.

DerDakon: I have updated the code, but have not tested it yet.

prot.c
DerDakon: The code only avoids the call to setgid() if setgroups() fails.

DerDakon:
ping

prot.c
leahneukirchen: I misread this. Not a fan of return on the same line. ;)

leahneukirchen (approved):

PR #158, opened 2020-06-18
make now() an inline function

DerDakon:
This is a thin wrapper about time(), so just let it do that as inline function. It can't be a macro, as it is called as a function but without arguments. Now it also returns time_t, which avoids losing precision on platforms where time_t is larger than long.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -818,10 +818,10 @@ strerr.h
 	./compile maildir.c
 
 maildir2mbox: \
-load maildir2mbox.o maildir.o prioq.o now.o myctime.o gfrom.o lock.a \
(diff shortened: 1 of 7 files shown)


DerDakon:
I just found it while looking at other things and decided that it's not worth to keep it because it has no advantage.

DerDakon:
Hm, noone seems to have spotted that it's not actually an inline function, but just a #define.

xenotrope:
A 64-bit time data type should be used everywhere. In looking at how prevalent the datetime handling code is in qmail, I found it to be sprawling and outdated. "Convert it to libtai" is much easier said than done, but replacing datetime_sec with a portable 64-bit TAI or TAI-like timestamp data type is a long-term solution that can be deployed in "substitute it where and when we find it" phases and stages.

PR #159, opened 2020-06-20
reintroduce alloc_re()

DerDakon:
This is used by the large-dns patch, and may be used by others. Add it back in a way that it will not be compiled in our normal build, but a simple compiler define allows external patches to find the needed environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:
   matrix:
     - BUILD_NAME="default"
     - BUILD_NAME="utmp" FORCE_UTMP="1" MAKEFLAGS="-o qtmp.h"
-    - BUILD_NAME="no_catman" AVOID_NROFF="1"
+    - BUILD_NAME="no_catman" AVOID_NROFF="1" EXTRA_CFLAGS="-DDONT_PROVIDE_OBSOLETE"
 compiler:
  - gcc
  - clang
@@ -14,7 +14,7 @@ addons:
     - pkg-config
     - check
 script:
- - echo "${CC} -O2 -Wall -Wshadow -Werror=implicit-function-declaration" > conf-cc
+ - echo "${CC} -O2 -Wall -Wshadow -Werror=implicit-function-declaration -Werror=deprecated-declarations ${EXTRA_CFLAGS}" > conf-cc
  - if [ -n "${FORCE_UTMP}" ]; then cp qtmp.h1 qtmp.h; fi
  - if [ -n "${AVOID_NROFF}" ]; then NROFF=true; export NROFF; fi
  - make ${MAKEFLAGS} it man
--- a/alloc.h
+++ b/alloc.h
@@ -6,4 +6,27 @@
 #define alloc(x) malloc(x)
 #define alloc_free(x) free(x)
 
+/*
+ * In-tree code must not call this deprecated function. It's here because
+ * some external patches still do. We intend to remove it as soon as is
+ * practical.
+ */
+#ifndef DONT_PROVIDE_OBSOLETE
+
+#if defined(__clang__) || defined(__GNUC__)
+#define _deprecated_ __attribute__((deprecated))
+#else
+#define _deprecated_
+#endif
+
+static inline int _deprecated_ alloc_re(void **x, unsigned int m, unsigned int n)
+{
+  void *y = realloc(*x, n);
+  (void)m;
+  if (y != NULL)
+    *x = y;
+  return !!y;
+}
+#endif
+
 #endif
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,7 +14,7 @@ task:
   pkginstall_script:
     - pkg install -y groff pkgconf check
   configure_script:
-    - echo "clang -O2 -Wall -Wshadow -Werror=implicit-function-declaration" > conf-cc
+    - echo "clang -O2 -Wall -Wshadow -Werror=implicit-function-declaration -Werror=deprecated-declarations" > conf-cc
   compile_script:
     - make ${MAKE_FLAGS} it man
   test_script:


schmonz:
> This is used by the large-dns patch

Tangent: assuming you mean [this one](https://www.ckdhr.com/ckd/qmail-103.patch), should we merge it? It’s been unconditionally applied in pkgsrc qmail for a long time.

DerDakon:
Yes, that one. Merging this should be considered.

DerDakon:
I reverted the #ifdef: it is now provided by default, but we can switch it off. I'm doing this in one CI build so we can detect if anything we add later would use it again.

Makefile
```diff
@@ -818,10 +818,10 @@ strerr.h
 	./compile maildir.c
 
 maildir2mbox: \
-load maildir2mbox.o maildir.o prioq.o now.o myctime.o gfrom.o lock.a \
```
schmonz, line 821: I don’t think this branch wants to include the `now()` change — rebase and re-push this one?

alloc.h
```diff
@@ -6,4 +6,19 @@
 #define alloc(x) malloc(x)
 #define alloc_free(x) free(x)
 
+/*
+ * This functions should not be used in any new code. It is provided so that
+ * external patches that rely on it's presence still work correctly.
+ */
+#ifndef DONT_PROVIDE_OBSOLETE
```
schmonz, line 13: Double-negative conditions are hard for my brain. What are our goals for people and code trying to interact with this function? I’m thinking once we ship 1.09 with intent-to-deprecate, we want

1. Zero callers in our tree
2. Not exposed by default, so we don’t start calling it by mistake
3. If someone has a patch that needs `alloc_re()`, they’ll have to `#define` something to get it

Before 1.09 ships, seems like it suffices to preserve the API as it was (and add a big warning comment like you’ve done) — unless you’re hoping to start accomplishing (2) now?

alloc.h
```diff
@@ -6,4 +6,19 @@
 #define alloc(x) malloc(x)
 #define alloc_free(x) free(x)
 
+/*
+ * This functions should not be used in any new code. It is provided so that
+ * external patches that rely on it's presence still work correctly.
```
schmonz, line 11: How about: “In-tree code must not call this deprecated function. It’s here because some external patches still do. We intend to remove it as soon as is practical.”

alloc.h
xenotrope: The only place I can see where DJB uses alloc_re() outside of gen_allocdefs.h in any of his projects is in env.c where it's used to add data into char **environ. Not only should manually calling alloc_re() be used sparingly, this patch and env.c may benefit from having their memory management logic revised. GEN_ALLOC_typedef() is used to provision more than just stralloc data structures, so I think that envalloc and resolvealloc typedefs may be suitable here just to standardize how notqmail handles dynamically allocating memory, unless someone thinks strallocs are a bad idea.

DerDakon: These places are already changed to use plain ```realloc()```.

Makefile
DerDakon: That's just the glitch of the GitHub Web UI as I rebased it on the remove-now branch before merging that to master, but at the same commit.

alloc.h
DerDakon: The intention here is indeed that we can build (and one CI run now does build) without that function being available, to enforce that we have no users in tree.

alloc.h
DerDakon: Done.

alloc.h
xenotrope: Glad to hear it. The question then is, apply the patch as is, apply the patch with realloc(), or standardize allocations globally and apply the patch with the standardized API?

DerDakon: If we are going to apply the large DNS patch (I think you mean that with "this patch"), then we will for sure fix it to use realloc() instead. This commit is needed for everything that may else be added by patch by users and which suddenly may fail to work otherwise.

schmonz: My preferred course, I think, is to start with this:

```c
#if defined(I_KNOW_THIS_DEPRECATED_CALL_WILL_GO_AWAY_SOON_AND_I_PROMISE_TO_FIX_MY_CALLING_CODE) \
    || !defined(REMOVE_DEPRECATED_CODE)
...
#endif 
```

(Ha-ha only serious about the first macro really being called that. That’s pretty much exactly what I did last time I had to migrate API callers, and it really did help people understand the implications of what they were doing.)

In the same PR branch, have TravisCI define `REMOVE_DEPRECATED_CODE` so we can start seeing how that goes.

Once 1.09 is released, remove the `|| !defined(REMOVE_DEPRECATED_CODE)` (and the now-superfluous TravisCI setting).

DerDakon: I just marked them formally as deprecated using compiler attributes and turned the usage of those functions to compiler errors in the CI. How about that?

schmonz: Yeah, that’s probably best. Does it give the build-time behaviors we want, if so, great.

.travis.yml
```diff
@@ -4,7 +4,7 @@ env:
   matrix:
     - BUILD_NAME="default"
     - BUILD_NAME="utmp" FORCE_UTMP="1" MAKEFLAGS="-o qtmp.h"
-    - BUILD_NAME="no_catman" AVOID_NROFF="1"
+    - BUILD_NAME="no_catman" AVOID_NROFF="1" EXTRA_CFLAGS="-DDONT_PROVIDE_OBSOLETE"
```
schmonz, line 7: Extraneous now?

DerDakon: It's a bit duplicate as it would fail because of -Werror anyway, but maybe just keep it out of paranoia.

PR #160, opened 2020-06-21
check presence of struct utimbuf and use it if present

DerDakon:
This fixes a compiler warning on most platforms:

```
qmail-send.c:471:21: warning: passing argument 2 of ‘utime’ from incompatible pointer type [-Wincompatible-pointer-types]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -662,6 +662,12 @@ trysgprm.c compile load
 	&& echo \#define HASSIGPROCMASK 1 || exit 0 ) > hassgprm.h
 	rm -f trysgprm.o trysgprm
 
+hasutbuf.h: \
+tryutbuf.c compile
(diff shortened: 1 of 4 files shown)


leahneukirchen:
utimes is POSIX.1-2001, perhaps just unconditionally use that?

DerDakon:
I thought about it, but it seems there were portability problems in ancient times. OTOH it is deprecated in newer versions of POSIX, so I think we are safer this way.

leahneukirchen:
I think every relevant platform will have it. It was in 4.2BSD and Linux libc4(!).

DerDakon:
Ok, the general opinion is to just use the struct unconditionally?

leahneukirchen:
No, I propose to use `utimes`, not `utime`.

DerDakon:
Version using ```utimes()``` implemented in #164.

PR #161, opened 2020-06-30
add missing includes for NULL

DerDakon:


DerDakon:
I messed up the Makefile for the ```waitpid()``` test so it failed, and then the code took the other branch in wait_nohang.c. I'm not aware of any other places, but there may be more hidden gems.

PR #162, opened 2020-07-01
qmail-remote: fix warnings about get() being used with wrong pointer signedness

DerDakon:
Became visible when building with clang.


DerDakon:
Right, but the functions it calls are expecting "char*"s.

PR #163, opened 2020-07-03
improve platform checks

DerDakon:


leahneukirchen:
I think all platforms we target have `<sys/select.h>`, so this can be used unconditionally.

DerDakon:
I strongly hope so, but for the moment we don't lose anything if we support also the old crap that doesn't.

josuah:
https://pubs.opengroup.org/onlinepubs/9699919799/functions/select.html
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_select.h.html

Implementations sometimes help operating systems libraries to comply with common practices.

PR #170, opened 2020-07-07
qmail-smtpd: allow [] to appear in received lines

DerDakon:
A client that can't find out the hostname should send "[ip.ad.dr.ess]" as HELO or EHLO string. But since the square brackets were considered unsafe qmail-smtpd mangled them to ? in the Received line actually written to the mail.


leahneukirchen:
I have a bunch of emails with `[]` in Received, I don't see how it does harm. :+1:

leahneukirchen:
sqmail also allows spaces, btw.

PR #171, opened 2020-07-22
get rid of error_* variables and just #define them to their errno equivalents

DerDakon:
One more part of #65.


leahneukirchen:
Adjust error_str.c and error_temp.c too?

DerDakon:
None of the extra cases in these files match with the values changed here. I have not touched any existing code that _uses_ these variables/defines.

DerDakon:
Fun fact: cdb_seek.c uses ```errno.h``` and ```EINTR``` and ```EIO``` without abstraction since qmail 1.01. As well as the things from ```seek.h```.

josuah (approved):
They are all in POSIX https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/errno.h.html and I should be present on everything notqmail targets.

mbhangui (approved):

PR #182, opened 2020-11-09
put all usernames into the same object

DerDakon:
Split out from #153. The .a version has the advantage that the linker can drop the strings it doesn't need. The .o version has all usernames in all affected binaries, but has much simpler build rules.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -706,25 +706,20 @@ install:
 	chmod 755 install
 
 instcheck: \
-load instcheck.o instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o uid.o \
-gid.o auto_usera.o auto_usero.o auto_userp.o auto_userq.o auto_userr.o \
-auto_users.o auto_groupq.o strerr.a substdio.a error.a str.a fs.a
+load instcheck.o instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o \
+ids.a strerr.a substdio.a error.a str.a fs.a
 	./load instcheck instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o \
-	uid.o gid.o auto_usera.o auto_usero.o auto_userp.o auto_userq.o \
-	auto_userr.o auto_users.o auto_groupq.o strerr.a substdio.a error.a \
-	str.a fs.a
+	ids.a strerr.a substdio.a error.a str.a fs.a
 
 instcheck.o: \
 compile instcheck.c strerr.h error.h readwrite.h hier.h
 	./compile instcheck.c
 
 instchown: \
-load instchown.o instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o uid.o \
-gid.o auto_usera.o auto_usero.o auto_userq.o auto_userr.o auto_users.o \
-auto_groupq.o strerr.a substdio.a error.a str.a fs.a
+load instchown.o instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o \
+ids.a strerr.a substdio.a error.a str.a fs.a
 	./load instchown instuidgid.o fifo.o hier.o auto_qmail.o auto_split.o \
-	uid.o gid.o auto_usera.o auto_usero.o auto_userq.o auto_userr.o \
-	auto_users.o auto_groupq.o strerr.a substdio.a error.a str.a fs.a
+	ids.a strerr.a substdio.a error.a str.a fs.a
 
 instchown.o: \
 compile instchown.c strerr.h error.h exit.h hier.h
@@ -1043,9 +1038,9 @@ qmail-control.9 conf-qmail conf-break conf-spawn
 
 qmail-getpw: \
 load qmail-getpw.o case.a substdio.a error.a str.a fs.a auto_break.o \
-auto_usera.o
+ids.a
 	./load qmail-getpw case.a substdio.a error.a str.a fs.a \
-	auto_break.o auto_usera.o 
+	auto_break.o ids.a
 
 qmail-getpw.0: \
 qmail-getpw.8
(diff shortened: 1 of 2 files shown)


mbhangui (approved):

schmonz:
Sorry it took me this long to notice, but these aren't IDs in the usual Unix numeric sense (and the sense in which pre-notqmail qmail was terribly annoying). They're user and group _names_. So maybe `usersgroups.o`? Or possibly `auto_usersgroups.o` since things generated with that mechanism currently get that prefix.

(Sorry.)

While I'm here, at a glance, most files that link `uid.o` also link `gid.o`. Maybe let's merge to `uidgid.o` (or `id.o`) as part of this PR?

(Double sorry.)

DerDakon:
The advantage of putting all these things in a .a would be that the resulting binaries will not grow. Adding uid.o and gid.o to that archive also is trivial then.

schmonz:
The users/groups and the corresponding numeric lookup routines do mostly get linked together. The `.o` make rules as presented here are nicely shrunken, but I think the savings from

1. Merging `uid.c` and `gid.c` into `uidgid.c` _and_
2. Linking `uidgid.o` into `auto_usergroup.a`

would be about as good. And then the `.a` link-time benefit you mention means we'd avoid any unwanted changes to binaries. So I think I finally have a strong opinion. 😜

DerDakon:
So it boils down to only have the first commit, and add uid.o and gid.o to the same archive, right?

schmonz:
> So it boils down to only have the first commit, and add uid.o and gid.o to the same archive, right?

Yes, I’ve finally landed on a preference for the approach in your first commit plus including those symbols in the same archive. Before you do that, try merging `uid.c` and `gid.c` into one `.c` file that generates one `.o` file, and then include that one additional `.o` in the archive. This would reduce some more duplicated make and C boilerplate without introducing any new symbols in binaries, if I understand correctly.

DerDakon:
It would introduce new symbols in qmail-rspawn and qmail-queue as they don't contain ```init_gid()``` yet. The linker can drop that only if it comes from it's own object file (not talking about -flto or advanced features like that).

schmonz:
Ah, okay. Then yes, exactly as you had suggested, please :-)

Makefile
```diff
@@ -1127,13 +1122,11 @@ qmail-log.5
 
 qmail-lspawn: \
 load qmail-lspawn.o spawn.o prot.o slurpclose.o coe.o sig.a wait.a \
-case.a cdb.a fd.a open.a stralloc.a substdio.a error.a str.a \
-fs.a auto_qmail.o uid.o gid.o auto_userp.o auto_userq.o auto_groupn.o \
-auto_spawn.o
+case.a cdb.a fd.a open.a stralloc.a ids.a substdio.a error.a str.a \
+fs.a auto_qmail.o
```
schmonz, line 1126: `ids.a` should still come after `auto_qmail.o`, and we lost the `auto_spawn.o` dependency

Makefile
```diff
@@ -1127,13 +1122,11 @@ qmail-log.5
...
 	./load qmail-lspawn spawn.o prot.o slurpclose.o coe.o sig.a wait.a \
-	case.a cdb.a fd.a open.a stralloc.a auto_qmail.o uid.o gid.o \
-	auto_userp.o auto_userq.o auto_groupn.o auto_spawn.o substdio.a \
-	error.a str.a fs.a
+	case.a cdb.a fd.a open.a stralloc.a auto_qmail.o \
```
schmonz, line 1128: `ids.a` between `auto_qmail.o` and `auto_spawn.o`

Makefile
```diff
@@ -1382,10 +1374,10 @@ error.h
 qmail-rspawn: \
 load qmail-rspawn.o spawn.o tcpto_clean.o coe.o sig.a open.a \
 lock.a wait.a fd.a stralloc.a substdio.a error.a env.a str.a \
-auto_qmail.o uid.o auto_userq.o auto_spawn.o
+auto_qmail.o auto_spawn.o ids.a
```
schmonz, line 1377: `ids.a` before `auto_spawn.o`

Makefile
```diff
@@ -1435,13 +1427,10 @@ qmail-send.service.in conf-qmail
...
 
 qmail-showctl: \
-load qmail-showctl.o uid.o gid.o auto_usera.o auto_userd.o auto_userl.o \
-auto_usero.o auto_userp.o auto_userq.o auto_userr.o auto_users.o auto_groupn.o \
-auto_groupq.o control.o open.a getln.a stralloc.a substdio.a error.a \
-str.a fs.a auto_qmail.o auto_break.o auto_patrn.o auto_spawn.o auto_split.o
-	./load qmail-showctl uid.o gid.o auto_usera.o auto_userd.o \
-	auto_userl.o auto_usero.o auto_userp.o auto_userq.o auto_userr.o \
-	auto_users.o auto_groupn.o auto_groupq.o control.o open.a getln.a \
+load qmail-showctl.o control.o open.a getln.a stralloc.a \
```
schmonz, line 1430: `ids.a` before `control.o` in dependency and build command

schmonz (changes requested):
Just these nits left to pick

schmonz (approved):
Retracting my remaining nitpicks about ordering. Having observed that replacing a list of `.o` leaves the new `.a` in between some remaining `.o`, you're moving the new `.a` near its siblings. I understand your intention now.

Separate commits would make this more apparent, but we've already workshopped this one more than enough! I agree that `git diff --word-diff` suffices to see what's happening here.

This PR is a terrific clarification of the make rules. It's so much easier to grok dependencies at a glance. qmail should have always looked up numeric IDs at runtime, and then it should always have been built this way.

josuah (approved):
This adds 0 portability cost and much readability.

PR #183, opened 2020-11-17
always include the own header first in all C files

DerDakon:
The idea is that the compiler will immediately error out if the header is not self contained, usually because of incomplete forward declarations in such headers.

### DIFF
--- a/cdbmss.c
+++ b/cdbmss.c
@@ -1,7 +1,8 @@
+#include "cdbmss.h"
+
 #include "readwrite.h"
 #include "seek.h"
 #include "alloc.h"
-#include "cdbmss.h"
 
 int cdbmss_start(c,fd)
 struct cdbmss *c;
--- a/coe.c
+++ b/coe.c
@@ -1,5 +1,6 @@
-#include <fcntl.h>
 #include "coe.h"
+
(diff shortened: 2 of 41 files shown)


schmonz (commented):
Certainly a style I’d not met before #182; thanks for proposing this more fully.

If this breaks patches, it seems likely to be trivial breakage, nothing that would require careful thought.

OTOH, maybe we could avoid any such problems while accomplishing the same goal (if I’ve understood it correctly) with a new build target that does a little `sed` (or whatever) to produce these edits and then calls the usual build target. And we could add this special build to our CI matrix, getting assurance that way until we live in a low-patch world. Possible? Sensible? Neither?

DerDakon:
Usually you want that when touching the header files, i.e. during development. Waiting for a CI target is IMHO a needless latency.

josuah (approved):
Compiles fine, ensure that no header miss from any .h, and that there is no difference between functions prototype and implementation.

mbhangui (approved):

schmonz (approved):
I've not seen this style before, but it makes sense, seems likely to help catch some mistakes, and doesn't seem likely to interfere much with patches. Looking forward to using it in practice.

PR #184, opened 2020-11-20
Fix 17 warnings from -Wparentheses.

schmonz:
The clang warning message says "using the result of an assignment as a
condition without parentheses". Parenthesize these assignments, because
their being also used as conditions is intentional.

This modifies qmail-remote:main() and qmail-smtpd:addrparse(), neither
of which are likely to affect users' patches.

### DIFF
--- a/auto-str.c
+++ b/auto-str.c
@@ -40,7 +40,7 @@ int main(int argc, char **argv)
   puts(name);
   puts("[] = \"\\\n");
 
-  while (ch = *value++) {
+  while ((ch = *value++)) {
     if (is_legible(ch)) {
       if (substdio_put(&ss1, &ch, 1) == -1)
         _exit(111);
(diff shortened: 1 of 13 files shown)


mbhangui (approved):

DerDakon:
I'm unsure if we may want to change that to something like this instead:

```
if ((dir = readdir(…)) != NULL)
```

schmonz:
I'm accustomed to reading these assignments as conditions, and somewhat prefer the brevity, but I'll willingly go along with any strong opinion anyone has about this.

josuah (approved):
DJB does not look like one to get a `=` vs `==` fail him, but I am not DJB, also, compilers are shouting at us nowadays.

DerDakon (approved):
Ok, the code has worked ever before. If some static analyzer is after us because we use assignments as boolean values that is another story. Let's silence the warnings that we already have now and care for the rest later.

PR #185, opened 2020-11-20
Fix 5 warnings from -Wdangling-else.

schmonz:
The clang warning message says "add explicit braces to avoid dangling
else". Brace ourselves.

This modifies qmail-inject:exitnicely() and qmail-local:checkhome(),
neither of which are likely to affect users' patches.

### DIFF
--- a/qmail-getpw.c
+++ b/qmail-getpw.c
@@ -37,7 +37,7 @@ int userext()
 	pw = getpwnam(username);
 	if (errno == error_txtbsy) _exit(QLX_SYS);
 	if (pw)
-	  if (pw->pw_uid)
+	  if (pw->pw_uid) {
 	    if (stat(pw->pw_dir,&st) == 0) {
 	      if (st.st_uid == pw->pw_uid) {
 		dash = "";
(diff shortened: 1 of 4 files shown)


mbhangui (approved):

mbhangui (approved):

qmail-getpw.c
```diff
@@ -47,6 +47,7 @@ int userext()
 	    }
 	    else
 	      if (error_temp(errno)) _exit(QLX_NFS);
+          }
```
DerDakon, line 50: The indentation here looks wrong.

DerDakon (changes requested):

qmail-getpw.c
schmonz: DJB's indentation was a mix of spaces and what seem to have been 8-space tabs. Check locally in your editor with brace-matching and it should look (somewhat) better.

When we eventually defeat patching, globally replacing tabs with spaces will give me a great deal of satisfaction.

DerDakon: Should be "Tab, Space, Space" before the brace then to match the if.

schmonz: Wow. Eagle eyes.

josuah (approved):
Also helps with readability on nested ifs

DerDakon (approved):

PR #186, opened 2020-11-20
Fix 3 warnings from -Wpointer-sign.

schmonz:
The clang warning message says "passing 'int *' to parameter of type
'socklen_t *' (aka 'unsigned int *') converts between pointers to
integer types with different sign". Indeed, the passed values are from
sizeof(). Make them unsigned.

### DIFF
--- a/tcp-env.c
+++ b/tcp-env.c
@@ -32,7 +32,7 @@ char temp[IPFMT + FMT_ULONG];
 
 int main(int argc, char **argv)
 {
- int dummy;
+ unsigned int dummy;
  char *proto;
  int opt;
  int flagremoteinfo;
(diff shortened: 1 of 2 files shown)


mbhangui (approved):

DerDakon (approved):
From accept(2):

>    The socklen_t type
>        In the original BSD sockets implementation (and on other older systems) the third argument of accept() was declared as an int *.  A POSIX.1g draft standard wanted to change it into a size_t *C; later POSIX standards and glibc 2.x have socklen_t * .

So, this was once right, but now the new type will cause less warning on recent (i.e. from this century) systems. This may then cause warnings on ancient systems, but who cares.

josuah (approved):
Maybe `socklen_t` instead of `unsigned int`? Not a huge difference, and better than plain int anyway.

[EDIT: multiple independent PoV converge]

DerDakon:
While I would use socklen_t in any of my code I fear that this is exactly that type of ancient compatibility issues that may actually break this for someone. When we eventually go and break everything we can break that, too ;)

PR #187, opened 2020-11-20
Fix 2 warnings from -Wincompatible-library-redeclaration.

schmonz:
The clang warning message says "incompatible redeclaration of library
function 'log2'". Call ours qslog2() instead, providing a compatibility
define for the old name in case patches need it.

### DIFF
--- a/qsutil.h
+++ b/qsutil.h
@@ -2,7 +2,16 @@
 #define QSUTIL_H
 
 extern void log1();
-extern void log2();
+extern void qslog2(char *, char *);
+
+#include "deprecated.h"
+#ifdef DEPRECATED_FUNCTIONS_AVAILABLE
+static inline void _deprecated_ log2(char *s1, char *s2)
+{
+  qslog2(s1,s2);
+}
+#endif
+
 extern void log3();
 extern void logsa();
 extern void nomem();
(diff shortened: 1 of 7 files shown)


mbhangui (approved):

qsutil.h
```diff
@@ -2,7 +2,10 @@
 #define QSUTIL_H
 
 extern void log1();
-extern void log2();
+extern void qslog2();
+#ifdef QMAIL_COMPAT_LOG2
```
DerDakon, line 6:
```c
#ifndef DONT_PROVIDE_OBSOLETE
```
as in ```alloc.h```?

schmonz: I knew we'd done this somewhere before and couldn't remember where!

qsutil.h
```diff
@@ -2,7 +2,11 @@
 #define QSUTIL_H
 
 extern void log1();
-extern void log2();
+extern void qslog2();
```
DerDakon, line 5: If you touch that line anyway, then please also provide the arguments in the prototype.

And when we already have _deprecated_ (which _I_ forgot about), then define log2() as an inline function that is annotated as deprecated, so it will spit warnings on use.

DerDakon (changes requested):

DerDakon (approved):

josuah (approved):

PR #188, opened 2020-11-20
Fix 1 warning from -Wempty-body.

schmonz:
Fix 1 warning from -Wempty-body.

The clang warning message says "if statement has empty body". Move the
statement inside the neighboring comment.


mbhangui (approved):

DerDakon (commented):
One could just turn that thing into a comment given that the instructions don't actually do anything.

josuah (approved):

DerDakon (approved):

PR #193, opened 2020-12-02
Fix CodeQL cpp/integer-multiplication-cast-to-long.

schmonz:
(Seen under the [project's "code scanning alerts"](https://github.com/notqmail/notqmail/security/code-scanning).)

"Multiplication result may overflow 'int' before it is converted to 'long'."

Recommendation: "Use a cast to ensure that the multiplication is done
using the larger integer type to avoid overflow."

### DIFF
--- a/qmail-send.c
+++ b/qmail-send.c
@@ -1043,7 +1043,7 @@ datetime_sec nextretry(birth,c)
 datetime_sec birth;
 int c;
 {
- int n;
+ datetime_sec n;
 
  if (birth > recent) n = 0;
  else n = squareroot(recent - birth); /* no need to add fuzz to recent */


qmail-send.c
```diff
@@ -1048,7 +1048,7 @@ int c;
  if (birth > recent) n = 0;
  else n = squareroot(recent - birth); /* no need to add fuzz to recent */
  n += chanskip[c];
- return birth + n * n;
+ return birth + (datetime_sec) n * n;
```
DerDakon, line 1051: Why didn't you just change the type of n to datetime_sec 4 lines above?

schmonz: Sure, I guess that makes more sense.

DerDakon (approved):

mbhangui (approved):

leahneukirchen (approved):

PR #195, opened 2020-12-05
Let QMAILQUEUE wrapper return custom error strings.

schmonz:
A common QMAILQUEUE use case is to prevent some messages from entering
the queue. We extend the qmail-queue(8) interface so that when rejecting
a message, a custom string can optionally be communicated to the caller
(for instance, qmail-smtpd). When this is desired, write to fd 6:

    "Dthis is a custom fatal error message"
      or
    "Zthis is a custom temporary failure message"

and exit 82.

This is the qmail-queue-custom-error-v2.netqmail-1.05.patch by Flavio
Curti <fcu-software@no-way.org>. It's been around since 2005 and
unchanged since 2007, when it switched from fd 4 to 6 in order to play
well with qmail-qfilter.

### DIFF
--- a/qmail-queue.8
+++ b/qmail-queue.8
@@ -81,6 +81,18 @@ Mail server permanently refuses to send the message to any recipients.
 .BR qmail-queue ,
 but can be used by programs offering the same interface.)
 .PP
+A special
+.B qmail-queue
+error code indicates a custom error string is written to descriptor 6:
+.TP 5
+.B 82
+Mail server refuses to send the message to any recipients:
+permanently for strings starting with "D",
+temporarily for strings starting with "Z".
+(Not used by
+.BR qmail-queue ,
+but can be used by programs offering the same interface.)
+.PP
 All other
 .B qmail-queue
 error codes indicate temporary errors:
--- a/qmail.c
+++ b/qmail.c
@@ -109,11 +109,11 @@ struct qmail *qq;
   if (!qq->flagerr) if (substdio_flush(&qq->ss) == -1) qq->flagerr = 1;
   close(qq->fde);
   substdio_fdbuf(&qq->ss,read,qq->fderr,qq->buf,sizeof(qq->buf));
-  while( substdio_bget(&qq->ss,&ch,1) && len < 255){
-    errstr[len]=ch;
+  while (substdio_bget(&qq->ss,&ch,1) && len < 255) {
+    errstr[len] = ch;
     len++;
   }
-  if (len > 0) errstr[len]='\0'; /* add str-term */
+  if (len > 0) errstr[len] = '\0'; /* add str-term */
 
   close(qq->fderr);
 
@@ -148,11 +148,10 @@ struct qmail *qq;
     case 81: return "Zqq internal bug (#4.3.0)";
     case 120: return "Zunable to exec qq (#4.3.0)";
     default:
-      if (exitcode == 82 && len > 2){
+      if (exitcode == 82 && len > 2)
         return errstr;
-      }
       if ((exitcode >= 11) && (exitcode <= 40))
-        return "Dqq permanent problem (#5.3.0)";
+	return "Dqq permanent problem (#5.3.0)";
       return "Zqq temporary problem (#4.3.0)";
   }
 }
--- a/qmail.c
+++ b/qmail.c
@@ -109,7 +109,6 @@ struct qmail *qq;
 {
   int wstat;
   int exitcode;
-  int match;
   static char errstr[256];
   size_t errlen;
 
--- a/qmail.c
+++ b/qmail.c
@@ -33,7 +33,6 @@ struct qmail *qq;
   if (pipe(pierr) == -1) {
     close(pim[0]); close(pim[1]);
     close(pie[0]); close(pie[1]);
-    close(pierr[0]); close(pierr[1]);
     return -1;
   }
  
--- a/qmail.c
+++ b/qmail.c
@@ -109,7 +109,7 @@ struct qmail *qq;
   if (!qq->flagerr) if (substdio_flush(&qq->ss) == -1) qq->flagerr = 1;
   close(qq->fde);
   substdio_fdbuf(&qq->ss,read,qq->fderr,qq->buf,sizeof(qq->buf));
-  while (substdio_bget(&qq->ss,&ch,1) && len < 255) {
+  while (substdio_get(&qq->ss,&ch,1) > 0 && len < 255) {
     errstr[len] = ch;
     len++;
   }
--- a/qmail.c
+++ b/qmail.c
@@ -38,9 +38,9 @@ struct qmail *qq;
  
   switch(qq->pid = fork()) {
     case -1:
-      close(pierr[0]); close(pierr[1]);
       close(pim[0]); close(pim[1]);
       close(pie[0]); close(pie[1]);
+      close(pierr[0]); close(pierr[1]);
       return -1;
     case 0:
       close(pim[1]);
--- a/qmail.c
+++ b/qmail.c
@@ -24,22 +24,32 @@ struct qmail *qq;
 {
   int pim[2];
   int pie[2];
+  int pierr[2];
 
   setup_qqargs();
 
   if (pipe(pim) == -1) return -1;
   if (pipe(pie) == -1) { close(pim[0]); close(pim[1]); return -1; }
+  if (pipe(pierr) == -1) {
+    close(pim[0]); close(pim[1]);
+    close(pie[0]); close(pie[1]);
+    close(pierr[0]); close(pierr[1]);
+    return -1;
+  }
  
   switch(qq->pid = fork()) {
     case -1:
+      close(pierr[0]); close(pierr[1]);
       close(pim[0]); close(pim[1]);
       close(pie[0]); close(pie[1]);
       return -1;
     case 0:
       close(pim[1]);
       close(pie[1]);
+      close(pierr[0]); /* we want to receive data */
       if (fd_move(0,pim[0]) == -1) _exit(120);
       if (fd_move(1,pie[0]) == -1) _exit(120);
+      if (fd_move(6,pierr[1]) == -1) _exit(120);
       if (chdir(auto_qmail) == -1) _exit(61);
       execv(*binqqargs,binqqargs);
       _exit(120);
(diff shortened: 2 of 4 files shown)


qmail.c
```diff
@@ -23,29 +23,40 @@ struct qmail *qq;
...
+  int pierr[2];
 
   setup_qqargs();
 
   if (pipe(pim) == -1) return -1;
   if (pipe(pie) == -1) { close(pim[0]); close(pim[1]); return -1; }
+  if (pipe(pierr) == -1) {
+    close(pim[0]); close(pim[1]);
+    close(pie[0]); close(pie[1]);
+    close(pierr[0]); close(pierr[1]);
```
DerDakon, line 35: This isn't needed: the creation of these fds failed, so there is nothing to close.

DerDakon (changes requested):
Please do the first commit using ``` git revert 5a214f931f1b5575e3cf4b037179eb5f665028c1```, and then just add your message on top.

If the patch is by Flavio then please just commit his patch as is using ```git commit --author ...``` to properly attribute him for the change, and then do your cleanups on to like you did.

schmonz:
I assumed we just lost `_bget()` in #84 and I missed it while reading. Phew. Will do.

qmail.c
schmonz: `pipe(2)` can't possibly half-succeed? For instance, create one, encounter a full file descriptor table, and fail the other?

DerDakon: I don't think so, that wouldn't make any sense and can't be properly detected.

qmail.c
```diff
@@ -88,10 +98,21 @@ struct qmail *qq;
...
   int exitcode;
+  char ch;
+  static char errstr[256];
+  size_t len = 0;
 
   qmail_put(qq,"",1);
   if (!qq->flagerr) if (substdio_flush(&qq->ss) == -1) qq->flagerr = 1;
   close(qq->fde);
+  substdio_fdbuf(&qq->ss,read,qq->fderr,qq->buf,sizeof(qq->buf));
+  while (substdio_bget(&qq->ss,&ch,1) && len < 255) {
```
DerDakon, line 109: We should clean this up a little bit, the ch variable isn't actually needed. One could pass ```errstr + len``` as second argument and just count len up.

DerDakon (changes requested):

qmail.c
schmonz: Like so?

```c
while (substdio_bget(&qq->ss,errstr+len++,1) && len < 256);
errstr[len] = '\0';
```

schmonz: Also, what's actually the difference between `substdio_bget()` and `substdio_get()`? I see lots of the latter reading a byte at a time and am wondering why we're different.

substdi.c
```diff
@@ -43,6 +43,17 @@ ssize_t substdio_feed(substdio *s)
   return r;
 }
 
+extern ssize_t substdio_bget(substdio *s, char *buf, size_t len)
```
DerDakon, line 46: No ```extern``` where the function is defined.

qmail.c
DerDakon: The length they pass to oneread() is the only real difference.

I would write the code as:

```c
while (substdio_bget(&qq->ss,errstr+len,1) && len < 256) {
  len++;
}
errstr[len] = '\0';
```

Otherwise len is one larger than before, and I see this as more readable.

substdi.c
schmonz: Oops

qmail.c
schmonz: Agreed. I think even better if I extract a little function.

There must be some a trick to express `substdio_get` in terms of `substdio_bget`, or vice versa, or at least have most of the work done by a static function they both call, but that doesn't need to be in this PR (or in 1.09).

DerDakon:
Would be nice if someone could actually test the new version of the patch in real life and see if it still works. Otherwise it looks fine to me.

schmonz:
Yep, will retest soon and report back.

schmonz:
Appears to work just as well as before. On notqmail 1.08 from pkgsrc on macOS Big Sur, with the original patch replaced by this one, I tested like so:

1. Appended `,QMAILQUEUE="/opt/pkg/bin/qmail-qfilter-queue",QMAILQUEUEFILTERS="control/localfilters"` to my `127.` tcprules and rebuilt the CDB
2. Created `control/localfilters` containing the line `/usr/local/bin/badrcptto.sh`
3. Created `/usr/local/bin/badrcptto.sh` with these contents:
```sh
#!/bin/sh

TOOBIG="xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
/bin/echo -n >&6 "Dsorry, friends (#5.7.1) ${TOOBIG}"
exit 82
```
4. Attempted, however halfheartedly, to send mail:
```
# mconnect
220 magnetic-babysitter.local ESMTP
mail me
250 ok
rcpt you
250 ok
data
354 go ahead
.
554 sorry, friends (#5.7.1) xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
quit
221 magnetic-babysitter.local
```

The message is rejected permanently, and is truncated (starting at "sorry") to 255 characters.

If I change `Dsorry` to `Zsorry`, the SMTP code is 451 instead.

mbhangui (approved):

DerDakon (approved):
What more can one expect ;)

qmail-queue.8
```diff
@@ -81,6 +81,18 @@ Mail server permanently refuses to send the message to any recipients.
 .BR qmail-queue ,
 but can be used by programs offering the same interface.)
 .PP
+A special
+.B qmail-queue
+error code indicates a custom error string is written to descriptor 6:
+.TP
```
leahneukirchen, line 87: Use `.TP 5` here, so it aligns with the other indents on the man page

qmail.c
```diff
@@ -83,15 +93,29 @@ void qmail_to(qq,s) struct qmail *qq; char *s;
   qmail_put(qq,"",1);
 }
 
+static size_t qmail_errstr(struct qmail *qq, char *s) {
+  size_t len = 0;
+  substdio_fdbuf(&qq->ss,read,qq->fderr,qq->buf,sizeof(qq->buf));
+  while (substdio_bget(&qq->ss,s+len,1) && len < 255) {
```
leahneukirchen, line 99: I'm not super familiar with substdio, but we do we need to read bytewise here and not all 255 bytes at once?

Also this loop won't terminate when substdio_bget returns -1 (on error).

schmonz: It does seem strange this way, and even if there’s a reason (maybe precisely because it doesn’t check for errors and assumes running out of bytes to read is the only error) I’m not sure why it needs `_bget` when there are plenty of 1-byte reads with `_get` sprinkled throughout the codebase.

Also good catch.

schmonz: On further review, the only difference between `substdio_get()` and `substdio_bget()` is that the latter might read fewer bytes at a time. This is superfluous when we’re reading a byte at a time.

For this PR, I’ll switch to one-byte `substdio_get()` with error checking to match how it’s being done everywhere else.

In a separate PR, I’ll bring back `_bget()`, mark it as deprecated, and attempt to highlight the very slight difference from `_get()`.

schmonz:
Retested and repushed.

leahneukirchen (approved):

schmonz:
We want to merge #199 before merging this, since the first commit for this PR (the original queue-custom-error patch) was using `substdio_bget()`.

qmail.c
```diff
@@ -24,29 +24,39 @@ struct qmail *qq;
...
   switch(qq->pid = fork()) {
     case -1:
       close(pim[0]); close(pim[1]);
       close(pie[0]); close(pie[1]);
+      close(pierr[0]); close(pierr[1]);
       return -1;
     case 0:
       close(pim[1]);
       close(pie[1]);
+      close(pierr[0]); /* we want to receive data */
```
josuah, line 48: We want to receive (read) data... and therefore we close the reading end (from where we read)... because now we are in the child and the child *sends* data?

If so, `/* we will send data to the parent */` would have gotten me faster to the point.  Maybe it is a fallacy of reading it as a patch and discovering it in-context makes sense though...

qmail.c
```diff
@@ -125,6 +149,8 @@ struct qmail *qq;
     case 81: return "Zqq internal bug (#4.3.0)";
     case 120: return "Zunable to exec qq (#4.3.0)";
     default:
+      if (exitcode == 82 && errlen > 2)
+        return errstr;
```
josuah, line 153: local buffer returned, but all fine: declared static.

josuah (commented):
Will come useful for extending qmail with error messages from the extensions code.

PR #196, opened 2020-12-05
Apply Christopher K. Davis's oversize DNS packet patch.

schmonz:
Originally named qmail-103.patch

Sometimes (rarely) the answer to a DNS query is larger than 512
bytes (the max that qmail allows (which was based on the UDP DNS
protocol definition (RFC 1025, section 4.2.1))). This is not a
widespread occurrence, yet, but can and does happen from time to
time. This patch allows DNS packets to be as big as the maximum
DNS response size, but does not waste memory if you never see
one that's that big.

### DIFF
--- a/dns.c
+++ b/dns.c
@@ -22,10 +22,12 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static union { HEADER hdr; unsigned char buf[PACKETSZ]; } response;
+static struct { unsigned char *buf; } response;
+static int responsebuflen = 0;
 static int responselen;
 static unsigned char *responseend;
 static unsigned char *responsepos;
+static u_long saveresoptions;
 
 static int numanswers;
 static char name[MAXDNAME];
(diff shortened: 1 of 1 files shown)


dns.c
```diff
@@ -22,10 +22,12 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static union { HEADER hdr; unsigned char buf[PACKETSZ]; } response;
+static struct { unsigned char *buf; } response;
+static int responsebuflen = 0;
 static int responselen;
 static unsigned char *responseend;
 static unsigned char *responsepos;
+static u_long saveresoptions;
```
DerDakon, line 30: plain unsigned long, I wonder why this compiles at all

dns.c
```diff
@@ -46,18 +48,33 @@ int type;
  errno = 0;
  if (!stralloc_copy(&glue,domain)) return DNS_MEM;
  if (!stralloc_0(&glue)) return DNS_MEM;
- responselen = lookup(glue.s,C_IN,type,response.buf,sizeof(response));
+ if (!responsebuflen)
+  if (response.buf = (unsigned char *)alloc(PACKETSZ+1))
```
DerDakon, line 52: Will likely warn because it needs an extra pair of (). And as this is considered "new" code it should use malloc() directly and doesn't need the cast then.

dns.c
```diff
@@ -46,18 +48,33 @@ int type;
...
+   responsebuflen = PACKETSZ+1;
+  else return DNS_MEM;
+
+ responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
+ if ((responselen >= responsebuflen) ||
+     (responselen > 0 && (((HEADER *)response.buf)->tc)))
+  {
+   if (responsebuflen < 65536)
+    if (alloc_re(&response.buf, responsebuflen, 65536))
```
DerDakon, line 61: This warns because we deprecated alloc_re().

dns.c
```diff
@@ -22,10 +22,12 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static union { HEADER hdr; unsigned char buf[PACKETSZ]; } response;
+static struct { unsigned char *buf; } response;
```
DerDakon, line 25: What about something like

```c
struct resp {
  HEADER hdr;
  char *buf;
} *response;
response = malloc(256);
```

Then we would not need to cast things to HEADER later. And maybe we could just do it in one line so the case that this has size 0 doesn't ever happen.

DerDakon (changes requested):
This should come in 2 commits: one applying the original patch with the proper author set, and one which adapts it to where we want to go with new code.

dns.c
schmonz: Maybe it doesn’t? This was an extremely low-effort initial submission on my part, just seeing the branch already exists and creating a PR. If it’s probably agreeable for 1.09 then I’ll make more effort, starting now :-)

schmonz: Comes from `<sys/types.h>`, at least on macOS

dns.c
schmonz: I haven't understood the original code well enough to try this. If you do, please push freely to #201.

dns.c
schmonz: Fixed in #201.

dns.c
schmonz: Fixed in #201.

dns.c
schmonz: Fixed in #201.

schmonz:
The named branch on this PR should remain as is, so I'm closing the PR. See #201 for the replacement PR.

PR #197, opened 2020-12-07
EAI support for qmail

mbhangui:
This adds EAI support to notqmail. Adapted from dapted from Arnt Gulbrandsen unicode address
support [here](http://arnt.gulbrandsen.priv.no/qmail/qmail-smtputf8.patch).

1. Use conf-smtputf8 to compile the EAI support
2. If conf-smtputf8 has -DSMTPUTF8, use tryidn2 to see if idn2_lookup_u8() from libidn2 can be used.
   if Yes, create hassmtputf8.h and #define SMTPUTF8
3. include hassmtputf8.h in qmail-smtpd.c, qmail-remote.c
4. The utf8read() function has been adapted from utf8received() by
   Erwin Hoffman
5. Following unit tests cases have been added
   - smtpcode() - in qmail-remote.c (test 3 digit codes, valid
     and junk codes)
   - mailfrom_parms() in qmail-smtpd.c (test SMTPUTF8 in the
     MAIL FROM parameter)
   - utf8read() in qmail-remote.c (test utf8, non-utf8)
   - get_capability() in qmail-remote.c to test for presence of EHLO
     capability

### DIFF
--- a/qmail-remote.c
+++ b/qmail-remote.c
@@ -27,6 +27,12 @@
 #include "timeoutconn.h"
 #include "timeoutread.h"
 #include "timeoutwrite.h"
+#include "hassmtputf8.h"
+#ifdef SMTPUTF8
+#include <idn2.h>
+#include "utf8read.h"
+#include "env.h"
+#endif
 
 #define HUGESMTPTEXT 5000
 
@@ -46,6 +52,13 @@ stralloc sender = {0};
 saa reciplist = {0};
 
 struct ip_address partner;
+
+#ifdef SMTPUTF8
+static stralloc idnhost = { 0 };
+static int      smtputf8 = 0; /*- if remote has SMTPUTF8 capability */
+static char    *enable_utf8 = 0; /*- enable utf8 */
+int             flagutf8;
+#endif
 
 void out(s) char *s; { if (substdio_puts(subfdoutsmall,s) == -1) _exit(0); }
 void zero() { if (substdio_put(subfdoutsmall,"\0",1) == -1) _exit(0); }
@@ -118,36 +131,50 @@ substdio smtpfrom = SUBSTDIO_FDBUF(saferead,-1,smtpfrombuf,sizeof(smtpfrombuf));
 
 stralloc smtptext = {0};
 
-static void get(unsigned char *uc)
+static void get1(unsigned char *uc)
 {
   char *ch = (char *)uc;
   substdio_get(&smtpfrom,ch,1);
-  if (*ch != '\r')
-    if (smtptext.len < HUGESMTPTEXT)
-     if (!stralloc_append(&smtptext,ch)) temp_nomem();
+  if (*ch != '\r' && smtptext.len < HUGESMTPTEXT &&
+      !stralloc_append(&smtptext,ch))
+    temp_nomem();
+}
+
+static unsigned long get3()
+{
+  char            str[4];
+  int             i;
+  unsigned long   code;
+
+  substdio_get(&smtpfrom,str,3);
+  str[3] = 0;
+  for (i = 0;i < 3;i++) {
+    if (str[i] == '\r') continue;
+    if (smtptext.len < HUGESMTPTEXT &&
+        !stralloc_append(&smtptext,str+i))
+      temp_nomem();
+  }
+  scan_ulong(str,&code);
+  return code;
 }
 
 unsigned long smtpcode()
 {
-  unsigned char ch;
-  unsigned long code;
+  unsigned char   ch;
+  unsigned long   code;
+  int             err = 0;
 
   if (!stralloc_copys(&smtptext,"")) temp_nomem();
-
-  get(&ch); code = ch - '0';
-  get(&ch); code = code * 10 + (ch - '0');
-  get(&ch); code = code * 10 + (ch - '0');
+  if ((code = get3()) < 200) err = 1;
   for (;;) {
-    get(&ch);
+    get1((char *)&ch);
+    if (ch != ' ' && ch != '-') err = 1;
     if (ch != '-') break;
-    while (ch != '\n') get(&ch);
-    get(&ch);
-    get(&ch);
-    get(&ch);
+    while (ch != '\n') get1((char *)&ch);
+    if (get3() != code) err = 1;
   }
-  while (ch != '\n') get(&ch);
-
-  return code;
+  while (ch != '\n') get1((char *)&ch);
+  return err ? 400 : code;
 }
 
 void outsmtptext()
@@ -211,6 +238,33 @@ void blast()
   substdio_flush(&smtpto);
 }
 
+#ifdef SMTPUTF8
+/*
+ * this function is general purpose
+ * we could use it when we add AUTH,
+ * STARTTLS, SIZE extensions
+ */
+int get_capability(const char *capa)
+{
+  int i = 0, len;
+
+  /* e.g.
+   * 250-PIPELINING
+   * 250-8BITMIME
+   * 250-SIZE 10000000
+   * 250-ETRN
+   * 250-STARTTLS
+   * 250-SMTPUTF8
+   * 250 HELP
+   */
+  len = str_len(capa);
+  for (i = 0; i < smtptext.len-len; ++i)
+    if (case_starts(smtptext.s+i,capa)) return 1;
+
+  return 0;
+}
+#endif
+
 stralloc recip = {0};
 
 void smtp()
@@ -221,14 +275,44 @@ void smtp()
  
   if (smtpcode() != 220) quit("ZConnected to "," but greeting failed");
  
+#ifdef SMTPUTF8
+  /*-
+   * this part of the code is general purpose
+   * it can be used for
+   * checking other extensions like
+   * AUTH, STARTTLS, SIZE, etc
+   */
+  substdio_puts(&smtpto,"EHLO ");
+  substdio_put(&smtpto,helohost.s,helohost.len);
+  substdio_puts(&smtpto,"\r\n");
+  substdio_flush(&smtpto);
+
+  if (smtpcode() != 250) { /*- do a HELO if EHLO fails */
+    substdio_puts(&smtpto,"HELO ");
+    substdio_put(&smtpto,helohost.s,helohost.len);
+    substdio_puts(&smtpto,"\r\n");
+    substdio_flush(&smtpto);
+
+    code = smtpcode();
+    if (code >= 500) quit("DConnected to "," but my name was rejected");
+    if (code != 250) quit("ZConnected to "," but my name was rejected");
+  } else /* EHLO succeeded. Let's check SMTPUTF8 capa */
+    smtputf8 = get_capability("SMTPUTF8"); /*- did the remote server advertize SMTPUTF8 */
+#else
   substdio_puts(&smtpto,"HELO ");
   substdio_put(&smtpto,helohost.s,helohost.len);
   substdio_puts(&smtpto,"\r\n");
   substdio_flush(&smtpto);
   if (smtpcode() != 250) quit("ZConnected to "," but my name was rejected");
+#endif
  
   substdio_puts(&smtpto,"MAIL FROM:<");
   substdio_put(&smtpto,sender.s,sender.len);
+#ifdef SMTPUTF8
+  if (enable_utf8 && (flagutf8 || utf8read()))
+    substdio_puts(&smtpto,"> SMTPUTF8\r\n");
+  else
+#endif
   substdio_puts(&smtpto,">\r\n");
   substdio_flush(&smtpto);
   code = smtpcode();
@@ -262,6 +346,9 @@ void smtp()
   if (code >= 500) quit("D"," failed on DATA command");
   if (code >= 400) quit("Z"," failed on DATA command");
  
+#ifdef SMTPUTF8
+  if (enable_utf8 && header.len) substdio_put(&smtpto, header.s, header.len);
+#endif
   blast();
   code = smtpcode();
   flagcritical = 0;
@@ -276,7 +363,12 @@ stralloc canonbox = {0};
 void addrmangle(stralloc *saout, char *s)
 {
   int j;
- 
+
+#ifdef SMTPUTF8
+  if (enable_utf8 && !flagutf8)
+    flagutf8 = containsutf8(s,str_len(s));
+#endif
+
   j = str_rchr(s,'@');
   if (!s[j]) {
     if (!stralloc_copys(saout,s)) temp_nomem();
@@ -325,6 +417,9 @@ int main(int argc, char **argv)
   if (chdir(auto_qmail) == -1) temp_chdir();
   getcontrols();
  
+#ifdef SMTPUTF8
+  enable_utf8 = env_get("UTF8");
+#endif
  
   if (!stralloc_copys(&host,argv[1])) temp_nomem();
  
@@ -342,6 +437,18 @@ int main(int argc, char **argv)
       relayhost[i] = 0;
     }
     if (!stralloc_copys(&host,relayhost)) temp_nomem();
+#ifdef SMTPUTF8
+  } else
+  if (enable_utf8) {
+      char *asciihost = 0;
+      if (!stralloc_0(&host)) temp_nomem();
+      switch (idn2_lookup_u8(host.s,(uint8_t**)&asciihost,IDN2_NFC_INPUT)) {
+        case IDN2_OK:     break;
+        case IDN2_MALLOC: temp_nomem();
+        default:          perm_dns();
+      }
+      if (!stralloc_copys(&idnhost,asciihost)) temp_nomem();
+#endif
   }
 
 
@@ -361,7 +468,11 @@ int main(int argc, char **argv)
 
  
   random = now() + (getpid() << 16);
+#ifdef SMTPUTF8
+  switch (relayhost ? dns_ip(&ip,&host) : dns_mxip(&ip,enable_utf8 ? &idnhost : &host,random)) {
+#else
   switch (relayhost ? dns_ip(&ip,&host) : dns_mxip(&ip,&host,random)) {
+#endif
     case DNS_MEM: temp_nomem();
     case DNS_SOFT: temp_dns();
     case DNS_HARD: perm_dns();
--- a/qmail-smtpd.c
+++ b/qmail-smtpd.c
@@ -24,6 +24,7 @@
 #include "timeoutread.h"
 #include "timeoutwrite.h"
 #include "commands.h"
+#include "hassmtputf8.h"
 
 #define MAXHOPS 100
 unsigned int databytes = 0;
@@ -33,6 +34,9 @@ GEN_SAFE_TIMEOUTWRITE(safewrite,timeout,fd,_exit(1))
 
 char ssoutbuf[512];
 substdio ssout = SUBSTDIO_FDBUF(safewrite,1,ssoutbuf,sizeof(ssoutbuf));
+#ifdef SMTPUTF8
+int  smtputf8 = 0, flagutf8 = 0;
+#endif
 
 void flush() { substdio_flush(&ssout); }
 void out(s) char *s; { substdio_puts(&ssout,s); }
@@ -217,6 +221,29 @@ int flagbarf; /* defined if seenmail */
 stralloc mailfrom = {0};
 stralloc rcptto = {0};
 
+#ifdef SMTPUTF8
+stralloc mfparms = {0};
+void mailfrom_parms(const char *arg)
+{
+  int i;
+  int len;
+
+  len = str_len(arg);
+  mfparms.len=0;
+  i = byte_chr(arg,len,'>');
+  if (i > 4 && i < len) {
+    while (len) {
+      arg++; len--;
+      if (*arg == ' ' || *arg == '\0' ) {
+         if (case_starts(mfparms.s,"SMTPUTF8")) smtputf8 = 1;
+         mfparms.len=0;
+      } else
+        if (!stralloc_catb(&mfparms,arg,1)) die_nomem();
+    }
+  }
+}
+#endif
+
 void smtp_helo(arg) char *arg;
 {
   smtp_greet("250 "); out("\r\n");
@@ -224,7 +251,14 @@ void smtp_helo(arg) char *arg;
 }
 void smtp_ehlo(arg) char *arg;
 {
-  smtp_greet("250-"); out("\r\n250-PIPELINING\r\n250 8BITMIME\r\n");
+  smtp_greet("250-"); out("\r\n250-PIPELINING\r\n");
+#ifdef SMTPUTF8
+  if (env_get("UTF8")) {
+    flagutf8 = 1;
+    out("250-SMTPUTF8\r\n");
+  }
+#endif
+  out("250 8BITMIME\r\n");
   seenmail = 0; dohelo(arg);
 }
 void smtp_rset(arg) char *arg;
@@ -235,6 +269,9 @@ void smtp_rset(arg) char *arg;
 void smtp_mail(arg) char *arg;
 {
   if (!addrparse(arg)) { err_syntax(); return; }
+#ifdef SMTPUTF8
+  if (flagutf8) mailfrom_parms(arg);
+#endif
   flagbarf = bmfcheck();
   seenmail = 1;
   if (!stralloc_copys(&rcptto,"")) die_nomem();
@@ -310,7 +347,7 @@ int *hops;
         if (flagmaybex) if (pos == 7) ++*hops;
         if (pos < 2) if (ch != "\r\n"[pos]) flagmaybey = 0;
         if (flagmaybey) if (pos == 1) flaginheader = 0;
-	++pos;
+        ++pos;
       }
       if (ch == '\n') { pos = 0; flagmaybex = flagmaybey = flagmaybez = 1; }
     }
@@ -372,7 +409,11 @@ void smtp_data(arg) char *arg; {
   qp = qmail_qp(&qqt);
   out("354 go ahead\r\n");
  
+#ifdef SMTPUTF8
+  received(&qqt,smtputf8 ? "UTF8SMTP" : "SMTP",local,remoteip,remotehost,remoteinfo,fakehelo);
+#else
   received(&qqt,"SMTP",local,remoteip,remotehost,remoteinfo,fakehelo);
+#endif
   blast(&hops);
   hops = (hops >= MAXHOPS);
   if (hops) qmail_fail(&qqt);
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -9,10 +9,12 @@ default: it
 
 .PHONY: clean default it test
 
-TESTBINS = unittest_stralloc unittest_blast unittest_prioq
+TESTBINS = unittest_stralloc unittest_blast unittest_prioq \
+		   unittest_utf8read
 
 clean:
-	rm -f $(TESTBINS) *.o blast.c
+	rm -f $(TESTBINS) *.o blast.c get_capability.c smtpcode.c \
+	mailfrom_parms.c
 
 it: $(TESTBINS)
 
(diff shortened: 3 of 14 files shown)


tests/get_function
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+#
+# Usage: get_function funcation_name filename
+# This script helps getting entire code snippet of a function
+# from the source code
+#
+start_line=`grep -n "^$1" $2|cut -d: -f1`
+sed -n "$start_line,$"p $2 | while IFS= read -r "line"
+do
+	echo "$line"
```
DerDakon, line 10: This will interpret e.g. \ and $ as shell characters and modify the contents. That's the reason for your compile error.

DerDakon (changes requested):

tests/get_function
mbhangui: This is a shell script. I have not understood why this is happening. The same Makefile and script is working for me.

mbhangui: Is there a way to pull in the libidn2-devel (libidn2-dev) library in the CI tests?

mbhangui: The logs show that the CI is using ubuntu bionic. The correct library to install for adding EAI support would be `libidn2-0-dev`

schmonz:
Thank you in particular for taking care to write tests for this functionality, especially since we've seen bugs in the currently available implementations.

When and how should this PR land? My thoughts: since SMTPUTF8 is a relatively small elaboration of qmail's existing 8-bit cleanliness, I'd like to see it built and enabled by default in some future notqmail release. Since the feature introduces a dependency on libidn2, I'd like to defer on-by-default until some later release that brings its own breaking changes. That leaves us merging EAI initially as off-by-default (and perhaps also autoconfigured-if-found). This can happen whenever the PR is ready, so it could potentially be part of 1.09 but shouldn't block that release if not.

mbhangui:
On Thu, 17 Dec 2020 at 02:21, Amitai Schleier <notifications@github.com>
wrote:

> When and how should this PR land? My thoughts: since SMTPUTF8 is a
> relatively small elaboration of qmail's existing 8-bit cleanliness, I'd
> like to see it built and enabled by default in some future notqmail
> release. Since the feature introduces a dependency on libidn2, I'd like to
> defer on-by-default until some later release that brings its own breaking
> changes. That leaves us merging EAI initially as off-by-default (and
> perhaps also autoconfigured-if-found). This can happen whenever the PR is
> ready, so it could potentially be part of 1.09 but shouldn't block that
> release if not.
>
It can be disabled by default by put a commenting out `-DSMTPUTF8` in
conf-smtputf8. But the man page will continue to describe the EAI feature.

qmail-remote.c
```diff
@@ -27,6 +27,12 @@
 #include "timeoutconn.h"
 #include "timeoutread.h"
 #include "timeoutwrite.h"
+#include "hassmtputf8.h"
+#ifdef SMTPUTF8
+#include <idn2.h>
+#include "utf8read.h"
+#include "env.h"
+#endif
```
schmonz, line 35: I'm seeing a lot of new ifdefs and it makes me a little uneasy about how "glanceable" our code is. We need some ifdefs for EAI because it's compile-time optional, but they can be (1) more isolated and (2) fewer in number. I'll give some examples of changes I'd like to see.

qmail-remote.c
```diff
@@ -47,6 +53,13 @@ saa reciplist = {0};
 
 struct ip_address partner;
 
+#ifdef SMTPUTF8
+static stralloc idnhost = { 0 };
+static int      smtputf8 = 0; /*- if remote has SMTPUTF8 capability */
+static char    *enable_utf8 = 0; /*- enable utf8 */
+int             flagutf8;
+#endif
+
```
schmonz, line 62: These variables don't cost much to have, so we could lose this ifdef. Another function of the ifdef is to document that these are EAI-specific; we could accomplish that by putting them in a new `.c` file (`eai_remote.c`?) and linking with the new `.o`.

qmail-remote.c
```diff
@@ -118,36 +131,50 @@ substdio smtpfrom = SUBSTDIO_FDBUF(saferead,-1,smtpfrombuf,sizeof(smtpfrombuf));
 
 stralloc smtptext = {0};
 
-static void get(unsigned char *uc)
```
schmonz, line 121: Haven't checked, just guessing we should probably retain the old name as a synonym in case patches call it.

qmail-remote.c
```diff
@@ -221,14 +275,44 @@ void smtp()
...
   if (smtpcode() != 250) quit("ZConnected to "," but my name was rejected");
+#endif
  
   substdio_puts(&smtpto,"MAIL FROM:<");
   substdio_put(&smtpto,sender.s,sender.len);
+#ifdef SMTPUTF8
+  if (enable_utf8 && (flagutf8 || utf8read()))
+    substdio_puts(&smtpto,"> SMTPUTF8\r\n");
+  else
+#endif
```
schmonz, line 315: It's cheap to check these variables even in the case where we built without EAI.

qmail-remote.c
```diff
@@ -262,6 +346,9 @@ void smtp()
   if (code >= 500) quit("D"," failed on DATA command");
   if (code >= 400) quit("Z"," failed on DATA command");
  
+#ifdef SMTPUTF8
+  if (enable_utf8 && header.len) substdio_put(&smtpto, header.s, header.len);
+#endif
```
schmonz, line 351: Same as previous

qmail-remote.c
```diff
@@ -276,7 +363,12 @@ stralloc canonbox = {0};
 void addrmangle(stralloc *saout, char *s)
 {
   int j;
- 
+
+#ifdef SMTPUTF8
+  if (enable_utf8 && !flagutf8)
+    flagutf8 = containsutf8(s,str_len(s));
+#endif
+
```
schmonz, line 371: Same as previous

qmail-remote.c
```diff
@@ -325,6 +417,9 @@ int main(int argc, char **argv)
   if (chdir(auto_qmail) == -1) temp_chdir();
   getcontrols();
  
+#ifdef SMTPUTF8
+  enable_utf8 = env_get("UTF8");
+#endif
```
schmonz, line 422: Same as previous. Regarding the environment variable, I'm familiar with setting env vars to control my locale settings, but is this an idiomatic way for a sysadmin to control mail server config? Assuming we're built with EAI, I'm wondering whether EAI should be enabled/disabled systemwide, per domain (incoming and outgoing), or some other level of granularity?

qmail-remote.c
```diff
@@ -342,6 +437,18 @@ int main(int argc, char **argv)
...
+  if (enable_utf8) {
+      char *asciihost = 0;
+      if (!stralloc_0(&host)) temp_nomem();
+      switch (idn2_lookup_u8(host.s,(uint8_t**)&asciihost,IDN2_NFC_INPUT)) {
+        case IDN2_OK:     break;
+        case IDN2_MALLOC: temp_nomem();
+        default:          perm_dns();
+      }
+      if (!stralloc_copys(&idnhost,asciihost)) temp_nomem();
+#endif
```
schmonz, line 451: This ifdef is a little extra tricky! It's only really needed around the `idn2_lookup_u8()` logic. That could be moved into `eai_remote.c` (along with an ifdef), wrapped in a descriptive function name, and called from here. Then we wouldn't need an ifdef here.

qmail-remote.c
```diff
@@ -361,7 +468,11 @@ int main(int argc, char **argv)
 
  
   random = now() + (getpid() << 16);
+#ifdef SMTPUTF8
+  switch (relayhost ? dns_ip(&ip,&host) : dns_mxip(&ip,enable_utf8 ? &idnhost : &host,random)) {
+#else
   switch (relayhost ? dns_ip(&ip,&host) : dns_mxip(&ip,&host,random)) {
+#endif
```
schmonz, line 475: Line 472 is enough all by itself :-)

qmail-smtpd.c
```diff
@@ -33,6 +34,9 @@ GEN_SAFE_TIMEOUTWRITE(safewrite,timeout,fd,_exit(1))
 
 char ssoutbuf[512];
 substdio ssout = SUBSTDIO_FDBUF(safewrite,1,ssoutbuf,sizeof(ssoutbuf));
+#ifdef SMTPUTF8
+int  smtputf8 = 0, flagutf8 = 0;
+#endif
```
schmonz, line 39: Let's just always define these, in a separate `eai_smtpd.c` if that helps readability.

qmail-smtpd.c
```diff
@@ -217,14 +221,44 @@ int flagbarf; /* defined if seenmail */
...
+      if (*arg == ' ' || *arg == '\0' ) {
+         if (case_starts(mfparms.s,"SMTPUTF8")) smtputf8 = 1;
+         mfparms.len=0;
+      } else
+        if (!stralloc_catb(&mfparms,arg,1)) die_nomem();
+    }
+  }
+}
+#endif
+
```
schmonz, line 246: So far, it appears the only use of this function (which I recognize from Arnt's patch) is to check whether the peer is trying to request `SMTPUTF8` for this delivery. Let's name it accordingly and mark it `static`. If we see other uses for it later, we can generalize then.

qmail-smtpd.c
```diff
@@ -217,14 +221,44 @@ int flagbarf; /* defined if seenmail */
...
 void smtp_ehlo(arg) char *arg;
 {
-  smtp_greet("250-"); out("\r\n250-PIPELINING\r\n250 8BITMIME\r\n");
+  smtp_greet("250-"); out("\r\n250-PIPELINING\r\n");
+#ifdef SMTPUTF8
+  if (env_get("UTF8")) {
+    flagutf8 = 1;
+    out("250-SMTPUTF8\r\n");
+  }
+#endif
```
schmonz, line 260: Cheap to check whether the admin wants to advertise `SMTPUTF8` -- let's just always do it.

qmail-smtpd.c
```diff
@@ -235,6 +269,9 @@ void smtp_rset(arg) char *arg;
 void smtp_mail(arg) char *arg;
 {
   if (!addrparse(arg)) { err_syntax(); return; }
+#ifdef SMTPUTF8
+  if (flagutf8) mailfrom_parms(arg);
+#endif
```
schmonz, line 274: Boolean checks are cheap

qmail-smtpd.c
```diff
@@ -372,7 +409,11 @@ void smtp_data(arg) char *arg; {
   qp = qmail_qp(&qqt);
   out("354 go ahead\r\n");
  
+#ifdef SMTPUTF8
+  received(&qqt,smtputf8 ? "UTF8SMTP" : "SMTP",local,remoteip,remotehost,remoteinfo,fakehelo);
+#else
   received(&qqt,"SMTP",local,remoteip,remotehost,remoteinfo,fakehelo);
+#endif
```
schmonz, line 416: Line 413 is enough by itself :-)

qmail-remote.c
DerDakon: if we #ifdef enable_utf8 to 0 then this comes for free as the compiler would optimize it away.

qmail-remote.c
DerDakon: And even if all of this is fine this variable name is a bit too generic.

qmail-remote.c
```diff
@@ -342,6 +437,18 @@ int main(int argc, char **argv)
       relayhost[i] = 0;
     }
     if (!stralloc_copys(&host,relayhost)) temp_nomem();
+#ifdef SMTPUTF8
+  } else
+  if (enable_utf8) {
+      char *asciihost = 0;
```
DerDakon, line 443: This should be NULL as it's a pointer variable

qmail-remote.c
DerDakon: And I would put it on a line distinct from the switch to make the code less convoluted.

qmail-smtpd.c
DerDakon: And every one on it's own line. Since they are global variables they are also automatically initialized to 0. Adding an explicit initializer is actually less efficient.

qmail-smtpd.c
```diff
@@ -217,14 +221,44 @@ int flagbarf; /* defined if seenmail */
 stralloc mailfrom = {0};
 stralloc rcptto = {0};
 
+#ifdef SMTPUTF8
+stralloc mfparms = {0};
+void mailfrom_parms(const char *arg)
+{
+  int i;
+  int len;
```
DerDakon, line 229: size_t

qmail-smtpd.c
```diff
@@ -217,14 +221,44 @@ int flagbarf; /* defined if seenmail */
...
+
+  len = str_len(arg);
+  mfparms.len=0;
+  i = byte_chr(arg,len,'>');
+  if (i > 4 && i < len) {
+    while (len) {
+      arg++; len--;
+      if (*arg == ' ' || *arg == '\0' ) {
+         if (case_starts(mfparms.s,"SMTPUTF8")) smtputf8 = 1;
```
DerDakon, line 238: I'm not sure if the case thing is actually right here. I have to reread the RfC, but I would bet this is not case insensitive.

DerDakon (changes requested):

qmail-smtpd.c
mbhangui: The same function can be used for getting the ESMTP AUTH and SIZE parameters. I will change this to static for the moment.

qmail-smtpd.c
mbhangui: Then let us remove flagutf8

qmail-smtpd.c
mbhangui: same as previous comment. Let us remove flagutf8

qmail-smtpd.c
mbhangui: RFC 5321 2.4.  General Syntax Principles and Transaction Model

Verbs and argument values (e.g., "TO:" or "to:" in the RCPT command and extension name keywords) are not case sensitive, with the sole exception in this specification of a mailbox local-part.
SMTP Extensions may explicitly specify case-sensitive elements. But RFC 6531 doesn't mention any case requirement for SMTPUTF8 keyword. So I assume it inherits the definition from RFC 5321

qmail-remote.c
mbhangui: Should SMTPUTF8 be ok instead of just UTF8. We can have a control file to have the setting per domain. We could add a line in getcontrols() to read a control file.

qmail-smtpd.c
mbhangui: I will do away with flagutf8.

qmail-smtpd.c
schmonz: If I understand what `flagutf8` is doing here, I'd keep it as a file global, making sure it gets set in exactly one place (probably in the same place where control files get read). I'd keep all the `flagutf8` checks, too -- just expressed in terms of the global boolean rather than the mechanics of how the config gets applied. And then I might call it `utf8_available`, or something that more precisely expresses this is whether the server offers the capability, as distinct from client-requested states like `utf8_requested` or `utf8_negotiated`.

If I don't understand what `flagutf8` is doing here, let's fix that first :-)

qmail-smtpd.c
schmonz: I was trying to say let's keep `flagutf8` and get rid of the `#ifdef`

mbhangui: Ok

josuah:
Should UTF-8 support everywhere really be an option? It is backward compatible with ASCII from the very byte level, so also being able to read UTF-8 might not break anyone's setup...

On the other hand, having the libidn2 dependency optional as we are pulling https://www.gnu.org/software/libunistring/ ?

My ideal would be to provide a tiny idn.c that implements punny code encoding (is that the aim?), but I yet have to provide this.

utf8read.c
```diff
@@ -0,0 +1,77 @@
...
+    if (r == -1) temp_read();
+
+    if (ch == '\n') {
+      if (!stralloc_append(&header, "\r")) temp_nomem(); /* received.c does not add '\r' */
+      if (!stralloc_append(&header, "\n")) temp_nomem();
+      if (case_starts(receivedline.s, "Date:")) return 0;  /* header to quit asap */
+      if (case_starts(receivedline.s, "Received: from")) received++;  /* found Received header */
+      if (received) {
+        for (i = 0; i < receivedline.len; i++)
+          if (*(receivedline.s + i) != ' ' && *(receivedline.s + i) != '\t')
```
josuah, line 48: I actually like this `*(variable + i)` notation without syntactic sugar for being very clear about what indices are, but the rest of the source goes with variable[i].

Might as well not introduce multiple coding style into a same daemon.

utf8read.c
```diff
@@ -0,0 +1,77 @@
...
+      }
+      if (!stralloc_copys(&receivedline, "")) temp_nomem();
+    } else {
+      if (!stralloc_append(&header, &ch)) temp_nomem();
+      if (!stralloc_catb(&receivedline, &ch, 1)) temp_nomem();
+    }
+  }
+  return 0;
+}
+#endif
```
josuah, line 77: The implementation from the patch was using a state machine, rather fun and creative, but plain dumb if/for code like yours feels more straightforward IMHO.

Here it is:

```
+void checkutf8message()
+{
+  int pos;
+  int i;
+  int r;
+  char ch;
+  int state;
+
+  if (containsutf8(sender.s, sender.len)) { utf8message = 1; return; }
+  for (i = 0;i < reciplist.len;++i)
+    if (containsutf8(reciplist.sa[i].s, reciplist.sa[i].len)) {
+      utf8message = 1;
+      return;
+    }
+
+  state = 0;
+  pos = 0;
+  for (;;) {
+    r = substdio_get(&ssin,&ch,1);
+    if (r == 0) break;
+    if (r == -1) temp_read();
+
+    if (!stralloc_append(&firstpart,&ch)) temp_nomem();
+
+    if (ch == '\r')
+      continue;
+    if (ch == '\t')
+      ch = ' ';
+
+    switch (state) {
+    case 6: /* in Received, at LF but before WITH clause */
+      if (ch == ' ') { state = 3; pos = 1; continue; }
+      state = 0;
+      /* FALL THROUGH */
+
+    case 0: /* start of header field */
+      if (ch == '\n') return;
+      state = 1;
+      pos = 0;
+      /* FALL THROUGH */
+
+    case 1: /* partway through "Received:" */
+      if (ch != "RECEIVED:"[pos] && ch != "received:"[pos]) { state = 2; continue; }
+      if (++pos == 9) { state = 3; pos = 0; }
+      continue;
+
+    case 2: /* other header field */
+      if (ch == '\n') state = 0;
+      continue;
+
+    case 3: /* in Received, before WITH clause or partway though " with " */
+      if (ch == '\n') { state = 6; continue; }
+      if (ch != " WITH "[pos] && ch != " with "[pos]) { pos = 0; continue; }
+      if (++pos == 6) { state = 4; pos = 0; }
+      continue;
+
+    case 4: /* in Received, having seen with, before the argument */
+      if (pos == 0 && (ch == ' ' || ch == '\t')) continue;
+      if (ch != "UTF8"[pos] && ch != "utf8"[pos]) { state = 5; continue; }
+      if(++pos == 4) { utf8message = 1; state = 5; continue; }
+      continue;
+
+    case 5: /* after the RECEIVED WITH argument */
+      /* blast() assumes that it copies whole lines */
+      if (ch == '\n') return;
+      state = 1;
+      pos = 0;
+      continue;
+    }
+  }
+}
```

Cannot most of it be summarized by `strncasestr(fieldstart, fieldlen, " with UTF8")` ? Case-insensitive searching  for the string " with utf8" within the "Received:" header.

josuah (changes requested):
Is there really any need to parse the "with UTF8SMTP" along with checking that the other host is supporting UTF8SMTP ?

If we have an email featuring UTF8 names and we need sending it we are screwed and end-up in the same result: the mail is not delivered. Does failing in a different way worth extra code?

And if we have UTF8 in a mail that lacks "with UTF8SMTP" in a `Received` header, shall we drop the mail, even though it shall also be well-encoded?

UTF8 email (and content in general) does not have problem with software without support for it (excepted for graphical programs that needs to fetch the codepoints), only with those that religiously reject content with the then-be-damned extra bit.

If anyone is having the a reference to some RFC for "with UTF8SMTP", I am highly interested. So far it looks like some indication for debugging which host scrambled the email rather than carrying any useful information.

Hopefully I am not misunderstanding everything.

josuah:
Thank you @mbhangui for bringing this starting point, that puts on the table everything to do.

mbhangui:
> Is there really any need to parse the "with UTF8SMTP" along with checking that the other host is supporting UTF8SMTP ?
> 
> If we have an email featuring UTF8 names and we need sending it we are screwed and end-up in the same result: the mail is not delivered. Does failing in a different way worth extra code?
> 

The relevant rfc is [RFC 6531 Section 3.6](https://tools.ietf.org/html/rfc6531)
My understading is that it is not necessary for the smtp client (qmail-remote). But the RFC doesn't make this mandatory.

  The MAIL command parameter SMTPUTF8 asserts that a message is an
   internationalized message or the message being sent needs the
   SMTPUTF8 support.  There is still a chance that a message being sent
   via the MAIL command with the SMTPUTF8 parameter is not an
   internationalized message.  An SMTPUTF8-aware SMTP client or server
   that requires accurate knowledge of whether a message is
   internationalized needs to parse all message header fields and MIME
   header fields [RFC2045] in the message body.  However, this
   specification does not require that the SMTPUTF8-aware SMTP client or
   server inspects the message.

> And if we have UTF8 in a mail that lacks "with UTF8SMTP" in a `Received` header, shall we drop the mail, even though it shall also be well-encoded?
> 

We shouldn't be dropping the mail. Rather qmail-remote shouldn't use the SMTPUTF8 extenstion in the MAIL FROM command

> UTF8 email (and content in general) does not have problem with software without support for it (excepted for graphical programs that needs to fetch the codepoints), only with those that religiously reject content with the then-be-damned extra bit.
> 
> If anyone is having the a reference to some RFC for "with UTF8SMTP", I am highly interested. So far it looks like some indication for debugging which host scrambled the email rather than carrying any useful information.
> 
RFC 6531 (latest) and RFC 5336 (old one)

> Hopefully I am not misunderstanding everything.

While going through the commands I realize that the original patch was written as per RFC 5336, hence the header with UTF8SMTP. RFC 5336 has been obsoleted by RFC6531. And the keyword UTF8SMTP should be changed to SMTPUTF8.
```
4.3.  WITH Protocol Types Sub-Registry of the Mail Transmission Types
      Registry

   IANA has modified or added the following entries in the "WITH
   protocol types" sub-registry under the "Mail Transmission Types"
   registry.

   +--------------+------------------------------+---------------------+
   | WITH         | Description                  | Reference           |
   | protocol     |                              |                     |
   | types        |                              |                     |
   +--------------+------------------------------+---------------------+
   | UTF8SMTP     | ESMTP with SMTPUTF8          | [RFC6531]           |
   | UTF8SMTPA    | ESMTP with SMTPUTF8 and AUTH | [RFC4954] [RFC6531] |
   | UTF8SMTPS    | ESMTP with SMTPUTF8 and      | [RFC3207] [RFC6531] |
   |              | STARTTLS                     |                     |
   | UTF8SMTPSA   | ESMTP with SMTPUTF8 and both | [RFC3207] [RFC4954] |
   |              | STARTTLS and AUTH            | [RFC6531]           |
   | UTF8LMTP     | LMTP with SMTPUTF8           | [RFC6531]           |
   | UTF8LMTPA    | LMTP with SMTPUTF8 and AUTH  | [RFC4954] [RFC6531] |
   | UTF8LMTPS    | LMTP with SMTPUTF8 and       | [RFC3207] [RFC6531] |
   |              | STARTTLS                     |                     |
   | UTF8LMTPSA   | LMTP with SMTPUTF8 and both  | [RFC3207] [RFC4954] |
   |              | STARTTLS and AUTH            | [RFC6531]           |
   +--------------+------------------------------+---------------------+
```

Good that you asked those questions and I see that there are minor changes to be made.

tests/unittest_utf8read.c
```diff
@@ -0,0 +1,389 @@
...
+  "  Ak+T6fp3gqhZszbocYfag1YGIDTrc8fxmjf9Ycmg5BvGzxa+cEVs7CYfJ9hFZM7m\n"
+  "  0AvoOdQ5fmWMCThNoHJzWDC3rTH7JFI=\n"
+  "Date: Thu, 3 Dec 2020 10:29:33 +0530\n"
+  "Message-ID: <20201203045933.3741.indimail@argos>\n"
+  "From: anonymous@argos.indimail.org\n"
+  "Subject: Output from your job      110\n"
+  "To: root\n"
+  "\n"
+  "error: can't create transaction lock on /var/lib/rpm/.rpm.lock (Resource temporarily unavailable)\n"
+  "error: /tmp/skype.gpgsig.zgIdKM: key 1 import failed.\n";
```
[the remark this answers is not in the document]
mbhangui: I will remove this. I just cut pasted the latest email in my Maildir/new folder without reliazing the skype thing.

tests/Makefile
```diff
@@ -62,3 +64,68 @@ unittest_prioq.o: \
...
+	(\
+	echo "#include \"stralloc.h\""; \
+	echo "#include \"str.h\""; \
+	echo "#include \"byte.h\""; \
+	echo "#include \"case.h\""; \
+	echo "extern stralloc mfparms;"; \
+	echo "extern int smtputf8;"; \
+	echo "extern void temp_nomem();"; \
+	./get_function "void mailfrom_parms" \
+	../qmail-smtpd.c|sed s}die_nomem}temp_nomem}g) > mailfrom_parms.c
```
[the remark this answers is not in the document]
mbhangui: Actually I don't like this. Looks very crude and can easily break when changes are mode to the original c code. But if it breaks, the test will fail during make itself and alert the maintainer to fix it.

mbhangui:
> While going through the commands I realize that the original patch was written as per RFC 5336, hence the header with UTF8SMTP. RFC 5336 has been obsoleted by RFC6531. And the keyword UTF8SMTP should be changed to SMTPUTF8.
> 
> ```
> 4.3.  WITH Protocol Types Sub-Registry of the Mail Transmission Types
>       Registry
> 
>    IANA has modified or added the following entries in the "WITH
>    protocol types" sub-registry under the "Mail Transmission Types"
>    registry.
> 
>    +--------------+------------------------------+---------------------+
>    | WITH         | Description                  | Reference           |
>    | protocol     |                              |                     |
>    | types        |                              |                     |
>    +--------------+------------------------------+---------------------+
>    | UTF8SMTP     | ESMTP with SMTPUTF8          | [RFC6531]           |
>    | UTF8SMTPA    | ESMTP with SMTPUTF8 and AUTH | [RFC4954] [RFC6531] |
>    | UTF8SMTPS    | ESMTP with SMTPUTF8 and      | [RFC3207] [RFC6531] |
>    |              | STARTTLS                     |                     |
>    | UTF8SMTPSA   | ESMTP with SMTPUTF8 and both | [RFC3207] [RFC4954] |
>    |              | STARTTLS and AUTH            | [RFC6531]           |
>    | UTF8LMTP     | LMTP with SMTPUTF8           | [RFC6531]           |
>    | UTF8LMTPA    | LMTP with SMTPUTF8 and AUTH  | [RFC4954] [RFC6531] |
>    | UTF8LMTPS    | LMTP with SMTPUTF8 and       | [RFC3207] [RFC6531] |
>    |              | STARTTLS                     |                     |
>    | UTF8LMTPSA   | LMTP with SMTPUTF8 and both  | [RFC3207] [RFC4954] |
>    |              | STARTTLS and AUTH            | [RFC6531]           |
>    +--------------+------------------------------+---------------------+
> ```

I misread the above. The protocol type still remans UTF8SMTP. RFC 5336 had just the description worded differently
```
   The "Mail Transmission Types" registry under the Mail Parameters
   registry is requested to be updated to include the following new
   entries:

   +---------------+----------------------------+----------------------+
   | WITH protocol | Description                | Reference            |
   | types         |                            |                      |
   +---------------+----------------------------+----------------------+
   | UTF8SMTP      | UTF8SMTP with Service      | [RFC5336]            |
   |               | Extensions                 |                      |
   | UTF8SMTPA     | UTF8SMTP with SMTP AUTH    | [RFC4954] [RFC5336]  |
   | UTF8SMTPS     | UTF8SMTP with STARTTLS     | [RFC3207] [RFC5336]  |
   | UTF8SMTPSA    | UTF8SMTP with both         | [RFC3207] [RFC4954]  |
   |               | STARTTLS and SMTP AUTH     | [RFC5336]            |
   +---------------+----------------------------+----------------------+
```

mbhangui:
> Is there really any need to parse the "with UTF8SMTP" along with checking that the other host is supporting UTF8SMTP ?
> 

I think I understand what you think is a problem. I'll explain below

> If we have an email featuring UTF8 names and we need sending it we are screwed and end-up in the same result: the mail is not delivered. Does failing in a different way worth extra code?
> 
> And if we have UTF8 in a mail that lacks "with UTF8SMTP" in a `Received` header, shall we drop the mail, even though it shall also be well-encoded?
> 

I see a problem. Let us say we are running SMTP service for clients to relay emails to external domains. qmail-smtpd advertises SMTPUTF8 and a client uses EAI and uses internationalized domain names in the FROM or the RCPT for a domain example.com (for example). The mail goes into the queue and qmail-remote finds that example.com doesn't support EAI. We should actually be dropping the mail with error like `553 server does not support internationalized email addresses`. If we had not advertised SMTPUTF8 in qmail-smtpd, the mail would have gone through.

One can argue that the original sender will get a bounce and then correct the message, but I'm not sure what we should do. Will read the RFC again to clear up this issue

PR #199, opened 2020-12-17
Restore (and deprecate) substdio_bget()

schmonz:


schmonz:
We want to merge this before merging #195, which is otherwise merge-ready.

substdi.c
```diff
@@ -43,12 +43,27 @@ ssize_t substdio_feed(substdio *s)
   return r;
 }
 
+#include "deprecated.h"
```
DerDakon, line 46: This header is not needed here: it is already included in substio.h, which holds the prototype for this function.

substdio.h
```diff
@@ -24,6 +24,10 @@ extern int substdio_bputs();
 extern int substdio_putsflush();
 
 extern ssize_t substdio_get(substdio *s, char *buf, size_t len);
+#include "deprecated.h"
```
DerDakon, line 27: Please move this header to the top where the other include is.

DerDakon (changes requested):

substdio.h
schmonz: I was aiming to follow the existing style, where (so far, anyway) we include the deprecation header just before it’s first used. I’m okay with changing the style. Should I also fix the other usages?

DerDakon: Uh, yes please. Why did I do this?

schmonz: It was probably me, that’s why :-)

schmonz: I've moved them all up, inside the include guards and, since it's the current style, after system headers (if any). #183 will want to move them slightly further up.

DerDakon (approved):

leahneukirchen (approved):

mbhangui (approved):

PR #201, opened 2020-12-24
Handle DNS packets up to max EDNS response size

schmonz:
Handle DNS packets up to max EDNS response size.

Original: qmail-103.patch

Sometimes (rarely) the answer to a DNS query is larger than 512 bytes. This is an important limit: it's the biggest response (unpatched) qmail can understand. This threshold was originally based upon the UDP DNS protocol definition (RFC 1025, section 4.2.1). It is still relatively unusual for DNS responses to be bigger than that limit, even though they are allowed to be. However, it does happen from time to time, particularly on domains with a lot of alternatives (if each MX record is approximately 50 bytes, you need 11 of them to cross the limit). This patch allows DNS packets to be as big as the maximum EDNS response size: 64kb (though RFC 6891 strongly recommends capping responses at 4kb).

(Description from http://www.memoryhole.net/qmail/#oversize-dns)

### DIFF
--- a/dns.c
+++ b/dns.c
@@ -49,16 +49,17 @@ int type;
  errno = 0;
  if (!stralloc_copy(&glue,domain)) return DNS_MEM;
  if (!stralloc_0(&glue)) return DNS_MEM;
- if (!responsebuflen)
+ if (!responsebuflen) {
   if ((response.buf = (unsigned char *)alloc(PACKETSZ+1)))
    responsebuflen = PACKETSZ+1;
   else return DNS_MEM;
+ }
 
  responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
  if ((responselen >= responsebuflen) ||
      (responselen > 0 && (((HEADER *)response.buf)->tc)))
   {
-   if (responsebuflen < 65536)
+   if (responsebuflen < 65536) {
     if (alloc_re(&response.buf, responsebuflen, 65536))
      responsebuflen = 65536;
     else return DNS_MEM;
@@ -66,6 +67,7 @@ int type;
     _res.options |= RES_USEVC;
     responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
     _res.options = saveresoptions;
+   }
   }
  if (responselen <= 0)
   {
--- a/dns.c
+++ b/dns.c
@@ -23,10 +23,12 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static union { HEADER hdr; unsigned char buf[PACKETSZ]; } response;
+static struct { unsigned char *buf; } response;
+static int responsebuflen = 0;
 static int responselen;
 static unsigned char *responseend;
 static unsigned char *responsepos;
+static u_long saveresoptions;
 
 static int numanswers;
 static char name[MAXDNAME];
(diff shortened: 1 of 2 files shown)


dns.c
```diff
@@ -47,18 +51,38 @@ int type;
  errno = 0;
  if (!stralloc_copy(&glue,domain)) return DNS_MEM;
  if (!stralloc_0(&glue)) return DNS_MEM;
- responselen = lookup(glue.s,C_IN,type,response.buf,sizeof(response));
+ if (!responsebuflen) {
+  if ((response.buf = (unsigned char *)alloc(PACKETSZ+1)))
```
DerDakon, line 55: replace that with malloc() as this is "new" code. The cast isn't needed then anymore as malloc() returns void*

DerDakon (changes requested):

dns.c
schmonz: Oops, I was sure I'd gotten that. Good catch. I've gone a little further and gotten rid of `alloc.h`.

DerDakon:
Funny that it's me, but… the last commit touches parts of the code for a cleanup not otherwise touched by this commit. I'm not sure if we want that at this stage.

schmonz:
Entirely fair. I don't see other patches touching `dns.c` in a quick grep of our qmail.org mirror, but the TLS patch definitely does. I'll save the wider cleanup for later.

DerDakon (approved):

mbhangui (approved):

dns.c
```diff
@@ -47,18 +51,38 @@ int type;
...
+ if (!responsebuflen) {
+  if ((response.buf = malloc(PACKETSZ+1)))
+   responsebuflen = PACKETSZ+1;
+  else return DNS_MEM;
+ }
+
+ responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
+ if ((responselen >= responsebuflen) ||
+     (responselen > 0 && (((HEADER *)response.buf)->tc)))
+  {
```
leahneukirchen, line 63: Why is this indented? Should merge with previous line.

DerDakon: because the line becomes too long for the ancient 80 char limits or so.

leahneukirchen: No? the line 60 is even longer. I'm talking about the `{` only.

leahneukirchen:
TBH I think in 2020 a static 64k buffer would solve this problem more easily.

schmonz:
> TBH I think in 2020 a static 64k buffer would solve this problem more easily.

Heh, could be. I guess with this PR I'm aiming for a combination of (1) avoiding original thought and (2) using the fix most familiar to longtime qmail users.

We're expecting to switch later to a better-matched DNS library (such as perhaps [s6-dns](https://skarnet.org/software/s6-dns/). So whatever we choose to do here, its days are numbered. Hopefully without needing bigint ;-)

dns.c
schmonz: This comes from the original patch. Looks like it was trying to match the existing brace style in that file, which at a glance was consistently weird in that way.

DerDakon:
I actually thought about using a "big" buffer, too. Maybe just submit that as an addon cleanup patch, @leahneukirchen?

dns.c
```diff
@@ -20,13 +20,17 @@ extern int res_search();
 #include "stralloc.h"
 #include "case.h"
 
+#define MAX_EDNS_RESPONSE_SIZE 65536
+
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static union { HEADER hdr; unsigned char buf[PACKETSZ]; } response;
+static struct { unsigned char *buf; } response;
```
josuah, line 28: Indeed, union does not support [flexible array member](https://en.wikipedia.org/wiki/Flexible_array_member).

What about degenerating it entirely to `char *` or even `void *` or even plain `HEADER`?

dns.c
```diff
@@ -47,18 +51,38 @@ int type;
  errno = 0;
  if (!stralloc_copy(&glue,domain)) return DNS_MEM;
  if (!stralloc_0(&glue)) return DNS_MEM;
- responselen = lookup(glue.s,C_IN,type,response.buf,sizeof(response));
+ if (!responsebuflen) {
+  if ((response.buf = malloc(PACKETSZ+1)))
+   responsebuflen = PACKETSZ+1;
+  else return DNS_MEM;
```
josuah, line 57: [#check-err-success]
This switches from checking for failure (`if (!(buf = malloc(4)) ERR;`) to checking for success, which conveniently looks like `try { ... } catch { ... }`, but adds more indentation depth, and makes switching just in one place could make failure handling more tedious to check through the source.

dns.c
```diff
@@ -47,18 +51,38 @@ int type;
...
+ if ((responselen >= responsebuflen) ||
+     (responselen > 0 && (((HEADER *)response.buf)->tc)))
+  {
+   if (responsebuflen < MAX_EDNS_RESPONSE_SIZE) {
+    unsigned char *newbuf = realloc(response.buf, MAX_EDNS_RESPONSE_SIZE);
+    if (newbuf) {
+     response.buf = newbuf;
+     responsebuflen = MAX_EDNS_RESPONSE_SIZE;
+    }
+    else return DNS_MEM;
```
josuah, line 70: same as [#check-err-success] above

dns.c
josuah: second thought, with a `static HEADER *responsebuf;`, responsebuf = malloc(PACKETSZ+1);` would not need cast anywhere (even in `lookup`, but just because it is not fully prototyped).

josuah (approved):

dns.c
schmonz: I generally prefer guard clauses too, for the reasons you state. The original patch gives us the code in this form, though, and although this particular clause would be simple enough to rephrase, I'm not generally feeling confident enough in my understanding of this code to refactor and re-test it. Mostly what I know about it is that I've been running it in production for a long time without apparent problems.

I'd be happy to see a followup PR from you or @leahneukirchen or @DerDakon or anyone else who feels more confident doing these cleanups :-)

dns.c
schmonz: Might be good. Happy to see it get tried :-)

PR #204, opened 2021-01-09
tcpto.c: initialize firstwhen variable

mbhangui:
Initialize firstwhen variable in tcpto.c and fix compiler warning

### DIFF
--- a/tcpto.c
+++ b/tcpto.c
@@ -74,6 +74,7 @@ void tcpto_err(ip,flagerr) struct ip_address *ip; int flagerr;
  char *record;
  datetime_sec when;
  datetime_sec lastwhen;
+ datetime_sec firstwhen = 0;
 
  if (!flagerr)
    if (!flagwasthere)
@@ -126,7 +127,6 @@ void tcpto_err(ip,flagerr) struct ip_address *ip; int flagerr;
  if (i >= n)
   {
    int firstpos = -1;
-   datetime_sec firstwhen;
    record = tcpto_buf;
    for (i = 0;i < n;++i)
     {


tcpto.c
```diff
@@ -74,6 +74,7 @@ void tcpto_err(ip,flagerr) struct ip_address *ip; int flagerr;
  char *record;
  datetime_sec when;
  datetime_sec lastwhen;
+ datetime_sec firstwhen = 0;
```
DerDakon, line 77: Why do you move the variable to a greater scope?

mbhangui: Just for readability to have all the three variables when, lasthwhen and firstwhen defined near to each other.

DerDakon: when you look at the code it is much easier to see that the variable is only used in the narrower scope, and it's closer to the code that actually uses it.

mbhangui:
have reduced the scope of the variable

moon-chilled:
The compiler warning is wrong—`firstwhen` will never be used uninitialized—and actually initializing `firstwhen` obscures that fact.

I think it would be better to refactor the whole loop, if you really want to avoid the warning; something like this maybe:

```c
if (i >= n && n == 0)
  i = -1;
else if (i >= n)
 {
  int firstpos;
  datetime_sec firstwhen;
  record = tcpto_buf;
  firstpos = 0;
  firstwhen = when = extract_record(&record);
  for (i = 1;i < n;++i)
   {
    when = extract_record(&record);
    if ((firstpos < 0) || (when < firstwhen))
     {
      firstpos = i;
      firstwhen = when;
     }
   }
  i = firstpos;
 }
```

PR #205, opened 2021-01-10
do not use Posix style -n argument for tail

DerDakon:
This was accidentially was changed from "tail -1" to "tail -n -1" instead of "tail -n 1". Even worse, the tail on Solaris doesn't seem to work with either options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,7 +115,7 @@ compile auto_usera.c
 
 auto_userd.c: \
 auto-str conf-users
-	./auto-str auto_userd `head -n 2 conf-users | tail -n -1` > auto_userd.c
+	./auto-str auto_userd `head -n 2 conf-users | tail -1` > auto_userd.c
 
 auto_userd.o: \
 compile auto_userd.c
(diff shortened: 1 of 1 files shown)


josuah (approved):
Solaris may accept `tail -n 1` or maybe `tail -n1` as per documentation :
https://docs.oracle.com/cd/E26502_01/html/E29030/tail-1.html

DerDakon:
> Solaris may accept `tail -n 1` or maybe `tail -n1` as per documentation :
> https://docs.oracle.com/cd/E26502_01/html/E29030/tail-1.html

Yes, but only for one of the 2 tail implementations. There seems to be only one head, with supports ```-n 1``` in contrast.

schmonz (approved):
I agree we need to fix the Solaris build, so I’m glad you set up an autobuild for it and found this.

The net change itself feels weird, calling `head` one way and `tail` another, but if that’s an accurate reflection of the situation then it probably doesn’t matter that it feels weird.

leahneukirchen (approved):

PR #208, opened 2021-01-24
style adjustment to dns.c, only the part that already changed

josuah:
- no need for a struct around remote.buf anymore, convert to plain (char *),
- check for failure (`if (!buf) ERR`) rather than success (`if (buf) ... else ERR`
  for consistency with surrounding code,
- small style changes.

### DIFF
--- a/dns.c
+++ b/dns.c
@@ -25,8 +25,8 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static struct { unsigned char *buf; } response;
-static int responsebuflen = 0;
+static char *responsebuf;
+static int responsebuflen;
 static int responselen;
 static unsigned char *responseend;
 static unsigned char *responsepos;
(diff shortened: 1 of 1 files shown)


dns.c
```diff
@@ -52,47 +52,44 @@ int type;
...
+    responsebuflen = MAX_EDNS_RESPONSE_SIZE;
     saveresoptions = _res.options;
     _res.options |= RES_USEVC;
-    responselen = lookup(glue.s,C_IN,type,response.buf,responsebuflen);
+    responselen = lookup(glue.s,C_IN,type,responsebuf,responsebuflen);
     _res.options = saveresoptions;
    }
   }
  if (responselen <= 0)
-  {
```
DerDakon, line 78: This line wasn't changed.

DerDakon:
ping?

mbhangui (approved):

josuah:
pong!
here we go: line was not changed and did change removed.

DerDakon (approved):

dns.c
```diff
@@ -25,8 +25,8 @@ extern int res_search();
 static unsigned short getshort(c) unsigned char *c;
 { unsigned short u; u = c[0]; return (u << 8) + c[1]; }
 
-static struct { unsigned char *buf; } response;
-static int responsebuflen = 0;
+static char *responsebuf;
```
DerDakon, line 28: You change this from ```unsigned char *``` to ```char *```. And at several places below this is assigned from an ```unsigned char *``` variable (when being allocated) or passed as parameter to ```res_query``` or ```res_search```, which expect an ```unsigned char *```. The latter is hidden because the ```lookup``` function pointer does not declare parameter types…

DerDakon (changes requested):

dns.c
josuah: Back to unsigned then. One example showing that expanding prototypes helps with collaboration.

DerDakon (approved):

Issue #203, opened 2021-01-08
Solaris make does not support "make -C"


DerDakon:
Tested it on a Solaris CI (coming soon). I'm not sure if we want to fix it, but I think it should not get lost.

PR #209, opened 2021-01-24
debug solaris makefile syntax issue

josuah:
any better way to debug it?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@ default: it
 
 .SUFFIXES: .0 .1 .5 .7 .8
 
+addresses.0: \
+addresses.5
+
 .1.0:
 	$(NROFF) -man $< >$@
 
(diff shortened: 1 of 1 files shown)


DerDakon:
Could you just turn this into a normal PR, changing only the Makefile and adding a comment about why this is needed? This could ease getting in other things, as more branches get affected by this.

DerDakon:
Doesn't help, see https://github.com/notqmail/notqmail/runs/1796785277?check_suite_focus=true

josuah:
I assumed I understood why it did work and attempted something that looks less weird.
I pushed the original version back but I do not understand what is going on.
Waiting code check and pinging back on IRC.

schmonz:
Do we remember what the error(s)/problem(s) looked like? In my testing for #231, I saw and fixed:

- `make man` getting an empty value of `NROFF`, despite `NROFF?=nroff`
- `make clean` and `make test` not working (#203)

Was there anything further?

schmonz:
This was almost certainly solved by #231. Please reopen (or file a new issue) if I missed something.

Issue #189, opened 2020-11-20
Default to -Wall (but not -Werror)


schmonz:
1. Get to warning-free builds in our usual dev environments

2. Submit a PR adding `-Wall -Werror` to `conf-cc` (and maybe shrinking our CI build matrix)

3. Test it on [every platform we can get our hands on](https://github.com/notqmail/notqmail/wiki/Platforms)

4. Fix whatever comes up

5. Merge

6. Rejoice!

7. Proceed directly to #48

schmonz:
#83

schmonz:
Must have click tapped something by mistake, this ain’t done yet

DerDakon:
Just to make this clear: this affects only CI builds, no? Otherwise I'll hate you for -Werror as it will break with every new compiler version and any released version.

schmonz:
Hmmm, I had been thinking we’d use it daily and fix whatever comes up, and ship with it and fix whatever users report. So yeah, the default `conf-cc`. Which scenario (or both) are you worried about?

DerDakon:
The point is to have it in the CI so we see everything that happens for us, but not have it in the default that ships to a user as that could easily break if it is built with a different compiler or on a different platform for no good reason. -Wall is fine, but not -Werror.

schmonz:
Heh, it keeps auto-closing this with _each_ linked issue instead of after _all_ of them (or not automatically at all). Guess I’m using this GitHub feature wrong.

I can enable `-Werror` in pkgsrc notqmail, pay attention to bulk builds on all platforms like I already do, and report here (and quickly work around there) whatever platform-specific warnings pop up in each notqmail release. This will give us pretty good platform coverage without setting a precarious default for everyone who builds.

schmonz:
For a data point, gcc 7.5.0 with `-DDEPRECATED_FUNCTIONS_REMOVED=1 -Wall -Werror -Wno-misleading-indentation -Wno-builtin-declaration-mismatch -Wno-pointer-sign -Wno-maybe-uninitialized -Wno-unused-variable` in `conf-cc` completes `make test` on macOS Big Sur.

schmonz:
Dang it, I keep forgetting that linking a PR to an issue makes the issue close when the PR does.

The approach in #211 (incrementally enable specific warnings in `conf-cc`, and maybe also specific warnings-as-errors) seems prudent, and also not mutually exclusive with other approaches (like enabling _all_ warnings and warnings-as-errors by default in our autobuilds).

[...]

PR #211, opened 2021-01-27
add noreturn attribute to a couple of functions

DerDakon:
When building with -Wextra there are a lot of false positives for fallthroughs, which happen because the compiler does not detect that these functions in fact never return. Add an attribute if building with gcc. According to the documentation it is supported since version 2.5, which is way older than anything one can reasonably expect this to work on.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -75,7 +75,7 @@ jobs:
         OS: ${{ matrix.host }}
 
     - name: configure
-      run: echo "${CC} -O2 -fPIC -Wall -Wshadow -Werror=implicit-function-declaration ${CFLAGS}" > conf-cc
+      run: echo "${CC} -O2 -fPIC -Wall -Wshadow -Werror=implicit-function-declaration -Wimplicit-fallthrough ${CFLAGS}" > conf-cc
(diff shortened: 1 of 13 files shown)


mbhangui (approved):

conf-cc
```diff
@@ -1,3 +1,3 @@
-cc -O2
+cc -O2 -Wimplicit-fallthrough
```
schmonz, line 1: I like this idea to incrementally enable specific warnings in `conf-cc` (all builds for all users) when we’re pretty sure we’ve thoroughly handled those warnings. Couldn’t we also enable `-Werror=implicit-fallthrough` here? (I might be making the same category of error as I was early in #189, but I don’t think so.)

DerDakon: Yes we should. And I should do it in the CI, not in the default file for everyone.

DerDakon: -Werror= doesn't work, there are other cases that still fallthrough.

schmonz (approved):

josuah (approved):
I am always sad to add compiler-specific flags and attributes.  But notqmail is developped for angry noisy compilers, so it is a necessary evil

PR #215, opened 2021-05-31
add more tests to prioq

tonytvo:
add some more tests for prioq

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -11,15 +11,22 @@ default: it
 
 TESTBINS = unittest_stralloc unittest_qmail-remote unittest_prioq unittest_qmail-send
 
+SINGLETESTBIN = unittest_prioq
+
 clean:
 	rm -f $(TESTBINS) *.o *-without-main.c
 
 it: $(TESTBINS)
 
+singleit: ${SINGLETESTBIN}
+
 test: it
 	@for tbin in $(TESTBINS); do \
 		./$$tbin || exit 1 ; \
 	done
+
+singletest: singleit
+	./${SINGLETESTBIN} || exit 1;
 
 unittest_stralloc: \
 ../load unittest_stralloc.o ../stralloc.a ../str.a ../error.a
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -61,7 +61,7 @@ unittest_prioq: \
 	`pkg-config --libs check`
 
 unittest_prioq.o: \
-../compile unittest_prioq.c ../prioq.h
+../compile unittest_prioq.c ../prioq.h ../gen_allocdefs.h
 	../compile unittest_prioq.c -I.. \
 	`pkg-config --cflags check`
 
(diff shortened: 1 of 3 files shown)


tests/Makefile
```diff
@@ -11,16 +11,23 @@ default: it
 
 TESTBINS = unittest_stralloc unittest_blast unittest_prioq
 
+SINGLETESTBIN = unittest_prioq
```
tonytvo, line 14: not sure if there's a way that I can run just a single test case and apparently, one option that I can see right now is to add target to run just a single specific test binary, it's better than run all of the test binary, but it still run all of the test cases in a single test binary

tests/unittest_prioq.c
```diff
@@ -1,10 +1,13 @@
 #include <check.h>
 
 #include <stdlib.h>
-
+#include <stdio.h>
 #include "prioq.h"
+#include "gen_allocdefs.h"
+
+GEN_ALLOC_readyplus(prioq,struct prioq_elt,p,len,a,100,check_readyplus)
```
tonytvo, line 8: I find it difficult to understanding the pre-processor that convert a specific function name to *_readyplus, so I try to write tests for it to understand it better

tests/Makefile
schmonz: `$ make test TESTBINS="unittest_stralloc unittest_prioq"`

(and `TESTBINS` can of course have just one value, too)

schmonz:
@tonytvo, I've rebased this onto `master` and had to force push (sorry). If you have any uncommitted or unpushed changes, let me know and I'll help you keep them. Otherwise, please `git fetch --all` and (assuming your name for my notqmail fork's remote is "schmonz"), on your branch, `git reset --hard schmonz/tvo-add-more-tests`. This will replace your local branch with what's on GitHub now.

tests/unittest_prioq.c
```diff
@@ -1,10 +1,24 @@
...
 
-START_TEST(test_prioq_empty)
+/* TODO
+- write tests for gen_alloc_readyplus
+- what's _internal?
+- what's ta stand for?
+- what's a good site for c language reference?
+- what's a good book to read more about c language?
+- what's a good code standard convention to follow?
+- what's a good way to run a single test case of single unit tests
```
josuah, line 16: - what's a good site for c language reference?
cppreference.com is covering plenty of C (and C++), but a good reference remains the C standard itself:
https://stackoverflow.com/questions/17014835/where-can-one-find-the-c89-c90-standards-in-pdf-format

- what's a good code standard convention to follow?
For a new project, I would advise either [Linux kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html) or [KNF](http://man.openbsd.org/style.9) (Kernel Normal Form) used by BSDs.
For an existing project, in particular if collaborating, I would preserve the original coding style of that project as much as possible. Which is a bit uncommon for qmail.

- what's a good way to run a single test case of single unit tests
If a test case takes too much time or produces too much output and you want to avoid it, you could split it away in another test file for instance.

DerDakon:
@schmonz maybe you need to move the branch over here or somewhere else? If you are going to open an new PR for the new location please link that here for the record.

PR #221, opened 2021-08-22
The pictures of how qmail forwards mail are plaintext and not markdown-compatible.

janicez:
I, ellenor@umbrellix.net, suggest creating doc/ and renaming PIC.(.*).md to PIC.\1.txt, as implemented in this pull request, so they're rendered properly in Github's text viewer (which interprets .md for Markdown).


DerDakon:
As the CI tells this breaks the installation as the files are not in the expected place anymore.

schmonz:
Ah yes, the install will need to be tweaked to continue finding the files in their new locations. This is probably best as separate PRs: one to rename the files that need renaming to avoid bad GitHub rendering, and another (if you're motivated to do it) to more generally move documentation files into a `doc/` subdir, which is a change we'll want to think about separately.

janicez:
Not mergeable.

PR #222, opened 2021-08-25
Remove TARGETS file

DerDakon:
As discussed in #217, this is for after 1.09.


josuah (approved):
This is more conventional, uses less command execution, and saves some ink too.

PR #223, opened 2021-08-27
add more _noreturn_ attributes as suggested my -Wsuggest-attribute=noreturn

DerDakon:
Since we have that attribute available anyway we can as well add it to more functions to allow the compiler to generate better code.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -14,7 +14,7 @@ task:
   pkginstall_script:
     - pkg install -y groff pkgconf check
   configure_script:
-    - echo "clang -O2 -Wall -Wshadow -Werror=implicit-function-declaration -Werror=deprecated-declarations" > conf-cc
+    - echo "clang -O2 -Wall -Wshadow -Werror=implicit-function-declaration -Werror=deprecated-declarations" -Wsuggest-attribute=noreturn > conf-cc
   compile_script:
     - make ${MAKE_FLAGS} it man
   test_script:
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -82,7 +82,7 @@ jobs:
         OS: ${{ matrix.host }}
 
     - name: configure
-      run: echo "${CC} -O2 -fPIC -Wall -Wshadow -Werror=implicit-function-declaration -Wimplicit-fallthrough ${CFLAGS}" > conf-cc
+      run: echo "${CC} -O2 -fPIC -Wall -Wshadow -Werror=implicit-function-declaration -Wimplicit-fallthrough -Wsuggest-attribute=noreturn ${CFLAGS}" > conf-cc
       env:
         CC: ${{ matrix.config.cc }}
         CFLAGS: ${{ matrix.config.cflags }}
--- a/forward.c
+++ b/forward.c
@@ -8,7 +8,7 @@
 
 #define FATAL "forward: fatal: "
 
-void die_nomem() { strerr_die2x(111,FATAL,"out of memory"); }
+void _noreturn_ die_nomem(void) { strerr_die2x(111,FATAL,"out of memory"); }
 
 struct qmail qqt;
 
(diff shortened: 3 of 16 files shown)


mbhangui:
one can use __STDC_VERSION__ in noreturn.h

https://stackoverflow.com/questions/28552467/is-there-a-way-to-use-gcc-attribute-noreturn-and-stdnoreturn-h-sanely-i

There is the suggestion to use `__attribute__((__noreturn__))` instead of `__attribute__((noreturn))` and also using __STDC_VERSION__ to avoid problems with including <stdnoreturn.h>

```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#define noreturn _Noreturn
#elif defined(__GNUC__)
#define noreturn __attribute__((noreturn))
#else
#define noreturn
#endif
```

DerDakon:
I wonder if we should not just use _Noreturn and add compatibility defines for that if it is not present.

leahneukirchen:
But why? If you define a macro anyway, just call it like the official C11 name.

DerDakon:
It's unclear to me what is the official name, if it is the define that comes from the header or the internal keyword. The plan is of course to use the standard one.

> This specifier is typically used through the convenience macro noreturn, which is provided in the header stdnoreturn.h
[CppReference.com](https://en.cppreference.com/w/c/language/_Noreturn)

This suggests that the ```noreturn``` define is not the real defined thing.

leahneukirchen:
The C11 keyword is `_Noreturn`, because for backwards compatibility they can't add new keywords with lowercase. But C11 also specifies the header `<stdnoreturn.h>` (7.23) which defines `noreturn` and doesn't want you make gouge your eyes out.

mbhangui:
including stdnoreturn.h breaks gcc if you use __attribute__ ((noreturn)). I had this problem last year compiling ezmlm-idx. ezmlm-idx uses this macro in die.h and cgi.h. I managed to find what the issue is. Here are two examples. The first example ignores noreturn because of stdnoreturn.h. Without stdnoreturn.h, it does the job. The second examples does the required job of adding noreturn attribute even with stdnoreturn.h included. The bug is mentioned [here](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=53182).
I faced this problem with gcc on fc33 and now on fc34 too (gcc version 11.2.1 20210728 (Red Hat 11.2.1-1) (GCC)).

** Example1 problem with including stdnoreturn.h

```
# cat a1.c
#include <unistd.h>
#include <stdnoreturn.h>
#define _noreturn_ __attribute__((noreturn))

__attribute__ ((__noreturn__)) void foo1()
{
        _exit(1);
}

_noreturn_ void foo2()
{
        _exit(1);
}
# gcc -Wsuggest-attribute=noreturn -c a1.c
a1.c:11:1: warning: ‘_Noreturn’ attribute directive ignored [-Wattributes]
   11 | {
      | ^
a1.c: In function ‘foo2’:
a1.c:10:17: warning: function might be candidate for attribute ‘noreturn’ [-Wsuggest-attribute=noreturn]
   10 | _noreturn_ void foo2()
      |                 ^~~~
```

Workaround is to use __attribute__ ((__noreturn__)) instead of __attribute__ ((noreturn))

** Example2

```
# cat a2.c
#include <unistd.h>
#include <stdnoreturn.h>
#define _noreturn_ __attribute__((__noreturn__))

__attribute__ ((__noreturn__)) void foo1()
{
        _exit(1);
}

_noreturn_ void foo2()
{
        _exit(1);
}
# gcc -Wsuggest-attribute=noreturn -c a2.c
```

DerDakon:
What works should be something like this:

```c
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#include <stdnoreturn.h>
#elif defined(__GNUC__)
#define noreturn __attribute__((__noreturn__))
#else
#define noreturn
#endif
```

josuah:
I felt that converting `_noreturn_` to `noreturn`, would go the same direction as converting `str_len` to `strlen`.

josuah (approved):
If there is anything happening about "noreturn" vs "_noreturn_", it would go in a separate commit?

All good with the `()` to `(void)` conversion as well.

PR #224, opened 2021-09-17
Add tests for qmail-send:job_*() functions.

samarjoe:
Like the qmail-remote:blast() tests, generate a copy of qmail-send.c
without main() for tests, which provide their own main(). Extract
job_*() declarations to qmail-send.h, also for use by tests.

This code contains contributions by 20 other participants from schmonz's
"Legacy Open Source Fridays" ensemble programming session, all of whom
are credited using GitHub's "Co-authored-by:"


DerDakon:
Another approach could be to not compile the qmail-send-without-main.c file into it's own object, but just #include it at the top of unittest_qmail-send.c. This avoids the need for a qmail-send.h. YMMV.

samarjoe:
> Another approach could be to not compile the qmail-send-without-main.c file into it's own object, but just #include it at the top of unittest_qmail-send.c. This avoids the need for a qmail-send.h. YMMV.

The header also extracted the jo struct

samarjoe:
> Another approach could be to not compile the qmail-send-without-main.c file into it's own object, but just #include it at the top of unittest_qmail-send.c. This avoids the need for a qmail-send.h. YMMV.

The header also extracted the jo struct into a single location for both the tests and the unittest_qmail-send to reference.

If we extracted that struct into the test program, then I think we could get rid of the qmail-send.h 

Just my feeling here, but I do like having a prototype into qmail-send for the tests.

DerDakon:
If you ```#include "qmail-send-without-main.c"``` you also have the struct definition from the original file.

samarjoe:
> 
> 
> If you `#include "qmail-send-without-main.c"` you also have the struct definition from the original file.

I went back and checked to see why we didn't do this at first, since I think it made a lot of sense. I then remembered, that including that .c caused a dependency conflict, specifically around the dependencies between check (mathcall log2) and qmail-send.c  log2). That may have been another reason we used a .h to link to a .o when creating these tests.

# make test
make[1]: Entering directory '/media/piExt/git/notqmail/samarjoe/schmonz/notqmail/tests'
../compile unittest_qmail-send.c -I.. \
`pkg-config --cflags check`
In file included from qmail-send-without-main.c:30,
                 from unittest_qmail-send.c:4:
../qsutil.h:10:33: warning: conflicting types for built-in function ‘log2’ [-Wbuiltin-declaration-mismatch]
 static inline void _deprecated_ log2(char *s1, char *s2)
                                 ^~~~
In file included from /usr/include/features.h:424,
                 from /usr/include/arm-linux-gnueabihf/sys/types.h:25,
                 from qmail-send-without-main.c:1,
                 from unittest_qmail-send.c:4:
/usr/include/arm-linux-gnueabihf/bits/mathcalls.h:133:1: error: conflicting types for ‘log2’
 __MATHCALL (log2,, (_Mdouble_ __x));
 ^~~~~~~~~~
In file included from qmail-send-without-main.c:30,
                 from unittest_qmail-send.c:4:
../qsutil.h:10:33: note: previous definition of ‘log2’ was here
 static inline void _deprecated_ log2(char *s1, char *s2)
                                 ^~~~
make[1]: *** [Makefile:93: unittest_qmail-send.o] Error 1
make[1]: Leaving directory '/media/piExt/git/notqmail/samarjoe/schmonz/notqmail/tests'
make: *** [Makefile:1881: test] Error 2

Is there some sort of compiler setting I'm not using to make the qsutils.h _deprecated_ log2 not conflict with mathcalls.log2?

Any suggestions on how to fix that?

DerDakon:
You could ```#define DEPRECATED_FUNCTIONS_REMOVED``` at the very beginning of the test source, ideally with a comment that one needs to remove log2() to make it compile, and then see if it works. If it still doesn't work not doing it that way is fine with me, even if I have a slight preference for that method as it means even less changes to qmail-send.c.

samarjoe:
> 
> 
> You could `#define DEPRECATED_FUNCTIONS_REMOVED` at the very beginning of the test source, ideally with a comment that one needs to remove log2() to make it compile, and then see if it works. If it still doesn't work not doing it that way is fine with me, even if I have a slight preference for that method as it means even less changes to qmail-send.c.

Thanks for the guidance @DerDakon , yes this worked beautifully. I have made the recommended changes and reworded the last commit to reflect.

josuah (approved):
Tests are also a good way to explain code.

This sounds like the best way to go: it would be easy to split job.c with job_*() later (once patch compatibility is not to consider).

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+
+extern struct job *jo;
+extern int numjobs;
+
+// START test helpers for the API we might prefer from qmail-send
+
+#define available 0
+#define unavailable 1
+
+static void job_init_this_many(int numslots) {
```
DerDakon, line 13: Please put the ```{``` for functions on a line of it's own as the other code also does.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+  numjobs = numslots;
+  job_init();
+}
+
+static void occupy_job(unsigned int index) {
+  jo[index].refs = unavailable;
+}
+
+static void occupy_all_jobs(void) {
+  for (int j = 0; j < numjobs; j++) {
```
DerDakon, line 23: You need to declare the variable on it's own as it's otherwise not valid C89 code. We have not yet raised the minimum required version to C99.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+
+START_TEST(test_qmail_job_init_sets_expected_default_state)
+{
+  ck_assert_ptr_null(jo);
+
+  job_init_this_many(4);
+
+  ck_assert_ptr_nonnull(jo);
+  ck_assert_uint_eq(numjobs, 4);
+  assert_job_at_index(0, available);
```
DerDakon, line 57: You set 4 jobs, but only check 2. Please check them all in a loop, and maybe also check that all other slots are unused.

Please always check all slots in all tests. This is unit tests, a bit of paranoia can't hurt here.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+  assert_job_avail(job_avail(), 1);
+}
+END_TEST
+
+START_TEST(test_qmail_job_open_slots_full)
+{
+  job_init_this_many(2);
+  occupy_all_jobs();
+
+  int index = job_open(1,2);
```
DerDakon, line 115: The variable must be declared at the start of the function, maybe add something like ```-std=c89``` to your conf-cc to catch this sort of compatibility issues.

Also add a space after comma.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+{
+  job_init_this_many(7);
+  assert_job_at_index(3, available);
+
+  job_close(3);
+  assert_job_at_index(3, -1);
+  job_close(3);
+  assert_job_at_index(3, -2);
+}
+END_TEST
```
DerDakon, line 175: I'm unsure about this test. It does test the current behavior, but as your comment shows you also see this as questionable. I would not test for this behavior as we may actually change it to something different, and suddenly having a failing test then could lead somebody to think that this behavior is actually wanted.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,239 @@
...
+  assert_job_at_index(index, unavailable);
+  job_close(index);
+  assert_job_at_index(index, available);
+  job_close(index);
+  ck_assert_int_lt(jo[index].refs, available);
+}
+END_TEST
+
+TCase
+*qmail_send_test_job_allocations(void)
```
DerDakon, line 198: I'm fine if you split the line for return value and function name, but the ```*``` is part of the return value then.

DerDakon (changes requested):
Sorry it took me so long.

tests/unittest_qmail-send.c
samarjoe: Thanks for the commentary here @DerDakon 

Wouldn't a subsequent PR that removes the behavior also include the rewritten, or newly written test?

I agree that having suddenly failing test could lead somebody to look into the test, but I also think that at some point this test would be updated to assert that the value will never be below zero.

I think that this characterization test is pretty useful since we can always see with certainty what the program is doing today, and we can determine if it should change and be more explicit about it's behavior in the future, for example, if available-slot (refs) is already set to 0 value (available), do not decrement, or explicitly set the value to 0, instead of a decrementing the value of refs in the function call.

tests/unittest_qmail-send.c
samarjoe: This makes, sense, and I do believe we originally were looping, but decided against it later on. @schmonz  maybe you can help recall why we retracted from that idea. I don't think there's anything wrong with putting `line 57` in a loop.

tests/unittest_qmail-send.c
samarjoe: This is good to know, thanks for this insight. I thought we were using C99.

tests/unittest_qmail-send.c
samarjoe: Can you help provide more guidance on the change you are expecting here? Apologies for the ignorance.

DerDakon: Either write

```c
TCase *qmail_send_test_job_allocations(void)
```

or

```c
TCase *
qmail_send_test_job_allocations(void)
```

tests/unittest_qmail-send.c
DerDakon: This is a counter overflow, and I can't think that it was intentional. When you do it like this you will run in odd behavior at other places, I have tested this while debugging another thing. The implicit assumption seems to be that this never underflows, which it doesn't if you do everything correct. You can see that other places e.g. check for ```refs != 0``` or ```!refs```, this was never intended for having anything negative.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,254 @@
...
+
+static void occupy_job(unsigned int index) 
+{
+  jo[index].refs = unavailable;
+}
+
+static void occupy_all_jobs(void) 
+{
+  int j;
+  for (j = 0; j < numjobs; j++)
```
DerDakon, line 28: Please add a newline after the variable just for consistency and easier reading.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,254 @@
...
+
+static void teardown(void) 
+{
+  numjobs = -88;
+}
+
+static void assert_all_slots_are_available(int numjobs) 
+{
+  int j;
+  for (j = 0; j < numjobs; j++)
```
DerDakon, line 62: Please add a newline after the variable just for consistency and easier reading.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,254 @@
...
+  release_job(numjobs - 1);
+
+  assert_job_avail(job_avail(), 1);
+}
+END_TEST
+
+START_TEST(test_qmail_job_open_slots_full)
+{
+  int index;	
+  job_init_this_many(2);
```
DerDakon, line 130: Please add a newline after the variable just for consistency and easier reading.

tests/unittest_qmail-send.c
```diff
@@ -0,0 +1,254 @@
...
+  numjobs = -88;
+}
+
+static void assert_all_slots_are_available(int numjobs) 
+{
+  int j;
+  for (j = 0; j < numjobs; j++) 
+  {
+    assert_job_at_index(j, available);
+    ck_assert_ptr_eq(jo[j].sender.s, 0);
```
DerDakon, line 65: should that be ```ck_assert_ptr_null()``` or ```, NULL```?

DerDakon (changes requested):
We are down to style nitpicks, I guess this is a good sign so I can eventually stop ;)

tests/unittest_qmail-send.c
samarjoe: I believe it's supposed to be a ck_assert_int_eq(jo[j].sender.s, 0);

the intention was to check whether other fields in the struct (besides refs was also 0'd out) as part of job_init().

I'll update this as well.

samarjoe: unittest_stralloc
Running suite(s): notqmail stralloc
100%: Checks: 2, Failures: 0, Errors: 0
unittest_qmail-remote
Running suite(s): notqmail qmail-remote blast
100%: Checks: 4, Failures: 0, Errors: 0
unittest_prioq
Running suite(s): notqmail prioq
100%: Checks: 6, Failures: 0, Errors: 0
unittest_qmail-send
Running suite(s): notqmail qmail-send job
SCHMONZ 1
SCHMONZ 2
SCHMONZ 3
SCHMONZ 4
90%: Checks: 10, Failures: 1, Errors: 0
unittest_qmail-send.c:67:F:test job allocations:test_qmail_job_init_sets_expected_default_state:0: Assertion 'jo[j].sender.s != NULL' failed: jo[j].sender.s == 0
SCHMONZ: would exit 1 here

Looks like after job_init() call the value for sender.s is 0. So we make that assertion to match as a characteristic test.

samarjoe: Yes, probably. I'll make that update.

DerDakon: Yes, but there is a difference beyond the standard. In gcc there is a ```__null``` builtin forever and that makes a difference e.g. when you have printf style and vargargs functions, as that 0 has always the size of a pointer which can make a difference when passed on the stack, and functions expecting a NULL terminator to a list of pointers will work correctly. For this the definition for gcc is ```#define NULL __null```, so one should use that constant whereever possible.

leahneukirchen: This makes no sense, and avoiding casts for sentinel values would actively break code on other compilers.

`__null` is only used on C++, and for C++11 there is an officall `nullptr` that should be used instead.

DerDakon: The point was to use ```NULL``` whereever possible and not ```0``` for a pointer value, even if the compiler tolerates that at a given point. For C++ one should always use ```nullptr``` if the compiler supports that. Regarding ``´__null``` you are right, that is only for C++ code. But using plain 0 instead of a proper NULL pointer constant could still mess up the argument length on the stack.

samarjoe:
> We are down to style nitpicks, I guess this is a good sign so I can eventually stop ;)

Speaking of style nitpicks, @tonytvo  and I realized that there were some whitespace at the end of some lines that needed some cleaning up. So we went ahead and made those changes and repushed the branch.

DerDakon:
The unittests look good so far.

@schmonz what about your Solaris debugging, have you found something that is actually needed? Can we get the actual fixes in their own PR and drop all these things from this PR please?

tests/unittest_qmail-send.c
leahneukirchen: Both will! You need `(char *)NULL` as a sentinel value for execvp etc.

DerDakon: The pain of variadic functions comes on top of the different (possible) size of ```0``` and ```NULL```, right.

schmonz:
I've extracted the Solaris build fixes and the Solaris and Ubuntu libcheck updates to a separate PR, #231. We should merge that first, then rebase this PR.

schmonz:
#231 is merged to master, so I've rebased this branch and force-pushed, leaving what is hopefully a sensible and focused commit history. @samarjoe @tonytvo assuming you don't have any additional local changes on your branch checkout, get rid of it and check it out freshly, something like so:

```sh
$ git checkout master
$ git branch -D samarjoe-test-qmail-send-jobs
$ git checkout samarjoe-test-qmail-send-jobs
```

But since you probably have more than one remote, that checkout might not quite work. Look in `git remote -v` for the remote that goes to `schmonz/notqmail.git` and checkout from that one, e.g.:

```sh
$ git checkout --track schmonz/samarjoe-test-qmail-send-jobs
```

There is probably an easier way and I'm probably about to learn it :-)

schmonz:
> There is probably an easier way and I'm probably about to learn it :-)

Much easier. Look in `git remote -v` for the right remote (say you've called it `schmonz`), and do this to replace your local branch with the remote's version of it:

```sh
$ git fetch schmonz
$ git reset --hard schmonz/samarjoe-test-qmail-send-jobs
```

samarjoe:
Is there anything left for this PR? Or are we looking good here?

DerDakon:
Ok, then I'll annoy everyone with commit cleanless now. The end result is quite good and I want to have that merged.

You create a qmail-send.h in the first commit, and then remove it again in the second one. I would prefer if you could squash at least that part of the second commit back into the first, so that no change has ever been done at the end.

While looking at the way you generate the qmail-send-without-main.c file I found a simpler way, which sadly doesn't work for the former blast.c as that would introduce additional dependencies there:

```shell
sed 's/^int main(/int orig_main(/' ../qmail-send.c > qmail-send-without-main.c
```

At least some parts of the third commit should also be squashed back into the first, i.e. just prevent introducing the coding style errors in the first place. At the end I suspect there will be barely anything left from that commit, but I have not inspected this closely.

The last commit needs a bit more context. Is there an uninitialized variable or why does Solaris fail? What does the XXX comment want to tell me in detail?

schmonz:
(With @samarjoe here) We decided to squash all the commits into one, keeping all the "Co-authored-by:". If we can't apply the simpler `sed` expression to the other test too, we'd like to defer that change to another time.

schmonz (approved):

DerDakon (approved):

PR #225, opened 2021-10-04
[WIP] Add a patch breakage checker to cirrus build

josuah:
I propose to move my https://github.com/josuah/notqmail-patch to https://github.com/notqmail/notqmail-patch (or any better name), so that anyone allowed to alter notqmail can also alter notqmail-patch.

Then, what about this to check for patch compatiblity ?

This also lets us maintain adapted versions of the patches, and adapt them as we break them.

In the end, some way for us to keep patches going as we move things around.

### DIFF
--- a/.github/workflows/patch.yml
+++ b/.github/workflows/patch.yml
@@ -1,0 +1,17 @@
+name: "Patch Compatibility"
+
+on:
+  push:
+    branches: [master, ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [master]
+jobs:
+  analyse:
+    name: Analyse
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout notqmail-patch
+      run: git clone https://github.com/josuah/notqmail-patch/
+    - name: Test whether we broke more patches than master did
+      run: make -C notqmail-patch BRANCH=${{ github.sha }} brokemaster


josuah:
On the above build, you see the "Patch Compatibility" at work on the 6th line.

I will try rebasing some patches in the meantime.

DerDakon:
Since cirrus-ci is often crowded and somewhat slow, can we get the same as a Github action? That would avoid any outside dependencies.

josuah:
My, my! It IS a Github action.  `.github/workflows/patch.yml` (title edited).

I first tried as a Cirrus build, not sure why, I'm new to CI.

DerDakon:
I think for a start: yes, please, move the repo into notqmail. But I dislike the "notqmail-patch" name, this can be understood as if it does something entirely different. How about "brokemaster" then?

josuah:
That looks like running to completion. We are good now, ready for review.

Let's see how it turns out in practice?
For now, the build is required, I can try to make it optional if you prefer.

DerDakon (changes requested):
LGTM, but I suggest also renaming the name of the workflow to "brokemaster" or something like that as well. Also please add a symlink to ```autobuilds/```.

josuah:
After too long, the suggested changes are completed.
Thank you for your patience.

DerDakon (approved):

Issue #210, opened 2021-01-26
when we broke the TLS patch, we didn't notice immediately


schmonz:
Sometime after 1.08, [the TLS patch](http://inoa.net/qmail-tls/) stopped applying to our tree. It's good we noticed this now so we can repair the situation before our next release. Some options:

1. Prepare a TLS patch rebased onto 1.09
2. Roll back our breaking changes and submit them as new PR(s) that document our known preconditions for merging
3. Coordinate with the TLS patch maintainer to publish a timely new version for notqmail 1.09 when we're ready to release

As usual, I'm extremely not excited about (1). We haven't maintained [our rebased patches](https://github.com/notqmail/notqmail/wiki/Patches) particularly well, and I don't blame us: they're easy to forget about, and if we do remember, the work isn't any fun. And users who are considering notqmail -- or already using notqmail and trying to track updates -- will much prefer being able to apply the patches they already have and know well. (I'm sure there'll be cases where we can't avoid providing rebased versions of known patches. Still, the fewer the better.)

(3) would be a pretty good outcome: minimal work for us, and users have already been trained to seek official updates to that patch over the years. If I were maintainer of a popular qmail patch, I'm not sure I'd be willing to make the effort to do this. But we can certainly ask. If he _is_ willing, the tradeoff here is that we've made the usability of our release depend on a third party. What if he changes his mind, or simply has no free time when we need him to?

My preference is for (2), especially since our breaking changes in this case appear to be small. For instance, we need to roll back calls to the `safe{read,write}` generator macros in `qmail-{remote,smtpd}.c` and restore the open-coded functions so that the TLS patch can modify them.

-----

Once we've preserved TLS for 1.09, we need to solve the meta-problem: in our current dev process, important patches can break and we might not know until too late.

[@josuah's auto-patcherator](https://notqmail.z0.is) is awesome, but we have to remember to look, and it seems to be the case that we haven't remembered. Again, I don't blame us. Our PR flow is already somewhat complicated.

Before we release 1.09, we need an automated way to be notified promptly that we broke a patch we intended not to break, so that we're sure not to release 1.09 with unintended breakage.

The cheapest time to avoid unintentional breakage, of course, is before merging a PR. But we already have to wait a while for the CI builds we've got. If that gets a lot longer, it'll be really tedious. The next-cheapest time to find out we broke something, then, is shortly after the merge.

[...]

PR #226, opened 2021-10-04
[WIP] Bump the smtp-tls patch branch to master

josuah:
As discussed in #210.

Another rebase was planned by @DerDakon for 1.09, this is an intermediate update to keep the patch abit more up to date. Or should we target the latest release rather than master?

Not for merging direcly: rather to push -f to the target branch for updating it.  I am dong a merge request for review purpose.


josuah:
IF you also want to drag in LibreSSL support to the TLS patch, one could try the below on an OpenSSL-based system and check whether this LibreSSL support broke OpenSSL support :

```
git checkout  josuah-patches/notqmail/smtp-libressl
make clean it
```

https://github.com/notqmail/notqmail/commit/54986362563a55e3741284b62aec0f826b0ac6e2

DerDakon:
I would do the rebase when 1.09 is out, and hopefully we can convince the patch author to do a new release around that timeframe, too, to reduce the diff.

Issue #147, opened 2020-06-10
Regression in 1.08: recipient-address qualification


schmonz:
The NetBSD script `/etc/daily` generates two emails to the sysadmin. It's called from `cron` like so:

    22 3 * * * /bin/sh /etc/daily 2>&1 | tee /var/log/daily.out | sendmail -t

In notqmail 1.07 (and netqmail before that), both emails were being addressed `To: root@my.fully.qualified.host.name`.

As of 1.08, one of the emails is instead addressed just `To: root`. Replacing `qmail-inject` with 1.07’s is sufficient to restore the previous behavior.

schmonz:
A minimal `qmail-inject-bug-reproducer`:

```sh
#!/bin/sh

AN_UNQUALIFIED_RECIPIENT=schmonz

host="notqmail-inject-test $(hostname)"
date="now $(date)"

echo "To: ${AN_UNQUALIFIED_RECIPIENT}"
echo "Subject: $host daily output for $date"
echo ""

echo "some daily insecurity output" | \
        mail -s "$host daily insecurity output for $date" ${AN_UNQUALIFIED_RECIPIENT}

echo "some daily output"
```

To run this the same way NetBSD's `/etc/daily` would:

    $ ./qmail-inject-bug-reproducer | sendmail -t

(It needs to be notqmail's `sendmail`, of course.)

This delivers me two messages, one of which has not qualified the `To:` address.

The one with the regression, injected via `mail` -> `sendmail` (options not yet known) -> `qmail-inject -a -- schmonz`:

```
Return-Path: <schmonz@schmonz.com>
Delivered-To: schmonz@miracle-dirt.schmonz.com
Received: (qmail 17228 invoked by uid 1000); 7 Dec 2020 15:28:08 -0000
Date: 7 Dec 2020 15:28:08 -0000
Message-ID: <20201207152808.3822.qmail@miracle-dirt.schmonz.com>
From: schmonz@schmonz.com
To: schmonz
Subject: notqmail-inject-test miracle-dirt.schmonz.com daily insecurity output for now Mon Dec  7 10:28:08 EST 2020

some daily insecurity output
```

And the one that does fully qualify the recipient, injected via `sendmail -t` -> `qmail-inject -H --`:

```
Return-Path: <schmonz@schmonz.com>
Delivered-To: schmonz@miracle-dirt.schmonz.com
Received: (qmail 7610 invoked by uid 1000); 7 Dec 2020 15:28:08 -0000
[...]

PR #229, opened 2021-12-23
qmail-inject: fix header parse regression

DerDakon:
When attempting to catch incorrectly formatted headers like "To: foo: bar, baz" to avoid errors when injecting bounces the host expansion of headers like "To: root" was also broken.

Instead of completely ignoring these lines when the recipient list was already given on the command line using -a option try to process these lines as before, but allow the parsing of these lines to silently fail. If a broken line is encountered this now simply does not reformat them to insert the hostname.

Bug introduced in commit 56e7c4a74087d47889355c2c55352cc4d0482020 in 1.08.

Fixes: #147

As I already said in that bug: I think we should just merge this without the tests so we can get a 1.09 out of the door. This fixes the regression, and when comparing the code before the breaking commit and this branch you see that it now only adds the "ignore errors" flag and routes the ```doordie()``` calls through a little helper that checks that flag. @schmonz verified that his usecase with the installed version now works.

### DIFF
--- a/qmail-inject.c
+++ b/qmail-inject.c
@@ -76,6 +76,14 @@ void doordie(sa,r) stralloc *sa; int r; {
  if (r == 1) return; if (r == -1) die_nomem();
  substdio_putsflush(subfderr,"qmail-inject: fatal: unable to parse this line:\n");
  substdio_putflush(subfderr,sa->s,sa->len); perm(); }
+/* call doordie, but if q is set ignore parse errors (i.e. r == 0) */
+static int doordie_rh(stralloc *sa, int r, int q)
+{
+  if (q && r == 0)
+    return 0;
+  doordie(sa, r);
+  return 1;
+}
 
 GEN_ALLOC_typedef(saa,stralloc,sa,len,a)
 GEN_ALLOC_readyplus(saa,stralloc,sa,len,a,10,saa_readyplus)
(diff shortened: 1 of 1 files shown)


DerDakon:
Includes the changes from #231 to make the CI green.

schmonz:
Could you try rebasing onto #232? Guessing it'll be easy and then we can be really sure.

schmonz:
Now that #232 is reliably failing all the autobuilds like it should, I tried rebasing this onto it. On my local macOS, the new tests pass. If they do likewise in all the autobuilds, then I'm indeed comfortable merging this _before_ #227 and #232.

schmonz:
[schmonz-Dakon-inject-fix-with-qfilter-and-tests](https://github.com/notqmail/notqmail/tree/schmonz-Dakon-inject-fix-with-qfilter-and-tests), a branch with this change rebased onto #232 (and therefore also #227), passes all our checks. We'd usually want to first merge the prerequisite-to-tests #227, then squash the tests from #232 onto this commit, then merge this. But #227 is sort-of-blocked on @bruceg issuing relicensed code. Having seen the tests pass is good enough for me for the moment, provided we don't forget to merge them as soon as we can. I approve this PR on its own, now.

schmonz (approved):

mbhangui (approved):

PR #231, opened 2022-01-13
autobuild updates

schmonz:
Fix the broken Solaris builds, including (I believe) solutions to #203 and #209.

As a prerequisite to #224, we need a newer libcheck on Solaris and Ubuntu. For the latter, we can get away with installing a binary package from a newer Ubuntu; for the former, we build from source. It might be worth following up this PR with one that [caches the built libcheck](https://github.com/actions/cache), as the Solaris builds are already very slow.

### DIFF
--- a/.github/workflows/solaris.yml
+++ b/.github/workflows/solaris.yml
@@ -10,21 +10,22 @@ on:
 jobs:
   build:
     name: Solaris-${{ matrix.config.name }}
-    runs-on: macos-latest
+    runs-on: macos-10.15
     env:
       CFLAGS: ${{ matrix.config.cflags }}
       NROFF: ${{ matrix.config.nroff }}
+      LIBCHECK_VERSION: "0.15.2"
     strategy:
       fail-fast: false
       matrix:
         config:
          - {
-           name: default,
+           name: default64,
            nroff: nroff,
            cflags: ""
          }
          - {
-           name: no_obsolete,
+           name: no_obsolete64,
            nroff: true,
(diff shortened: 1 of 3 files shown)


.github/workflows/solaris.yml
```diff
@@ -35,13 +35,21 @@ jobs:
...
+      uses: vmactions/solaris-vm@v0.0.5
       with:
         envs: 'CFLAGS NROFF'
-        prepare: pkgutil -y -i libcheck_dev gcc4core pkgconfig
+        prepare: |
+          pkgutil -y -i gcc4core pkgconfig curl
+          curl -L -O https://github.com/libcheck/check/releases/download/0.15.2/check-0.15.2.tar.gz
+          gunzip check-0.15.2.tar.gz
+          tar -xvf check-0.15.2.tar
+          cd check-0.15.2
```
DerDakon, line 46: Please put that version number in a variable so we can easily update it.

Makefile
```diff
@@ -1,7 +1,7 @@
 # Don't edit Makefile! Use conf-* for configuration.
 
 SHELL=/bin/sh
-NROFF?=nroff
+NROFF=nroff
```
DerDakon, line 4: It would be awesome if you could add to the comment fixing this a "Fixes: <sha1>" line with a reference to the commit that broke it. Same for the "make -C" ones.

DerDakon (changes requested):

DerDakon (approved):

mbhangui (approved):

leahneukirchen (approved):

PR #232, opened 2022-01-20
Test qmail-inject recipient address qualification.

### LINKED ISSUE #147: Regression in 1.08: recipient-address qualification

schmonz:
Using `qmail-qfilter` (proposed in #227), turn the minimal reproducer from #147 into an automated regression test. It's an integration test, not a unit test, because it executes `qmail-inject`[*] the same way our systems would. Thanks to `qmail-qfilter`, the test does not involve a real queue in any way.

#229 probably fixes #147, and these tests will help us be confident of that.

[*] Or rather, `tests/testable/qmail-inject`, which is identical modulo linking with a custom definition of `auto_qmail` that points into the source tree

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -7,19 +7,28 @@ SHELL=/bin/sh
 
 default: it
 
-.PHONY: clean default it test
+.PHONY: clean default it unittest integrationtest test
 
-TESTBINS = unittest_stralloc unittest_blast unittest_prioq
+UNITTESTS = unittest_stralloc unittest_blast unittest_prioq
+INTEGRATIONTESTS = integrationtest_inject
 
 clean:
-	rm -f $(TESTBINS) *.o blast.c
+	rm -f $(UNITTESTS) $(INTEGRATIONTESTS) *.o blast.c
+	rm -rf testable
 
-it: $(TESTBINS)
+it: $(UNITTESTS) $(INTEGRATIONTESTS)
 
-test: it
-	@for tbin in $(TESTBINS); do \
-		./$$tbin || exit 1 ; \
+unittest: $(UNITTESTS)
+	@for testprogram in $(UNITTESTS); do \
+		./$$testprogram || exit 1; \
 	done
+
+integrationtest: $(INTEGRATIONTESTS)
+	@for testprogram in $(INTEGRATIONTESTS); do \
+		./$$testprogram || exit 2; \
+	done
+
+test: unittest integrationtest
 
 unittest_stralloc: \
 ../load unittest_stralloc.o ../stralloc.a ../str.a ../error.a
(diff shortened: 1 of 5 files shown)


DerDakon:
Almost there ;)

I would very much prefer to not have to create a separate inject binary. Would it be possible to let the filter grep for the expected "To: " line and return success or failure? Then it should be possible to just qmail-inject and look at it's return code to see if the filter has accepted the mail or not. At least that would be the solution with the least possible side effects I can think of.

schmonz:
Yes, I'd feel much better if we could directly test the real `qmail-inject` we're about to install. Trouble is, before we've installed, `qmail-inject` hits `die_chdir()` -- and if we _have_ installed notqmail already, `/var/qmail` is still the wrong place to be looking. So at least in this one case so far, we need a way to override the value of `auto_qmail` for testing.

But how? Linker tricks aren't portable enough. Relinking a fake `auto_qmail` into the _real_ `qmail-inject` is a no-no (people need to be able to run the tests and then install). Maybe we introduce an environment variable for the purpose of overriding `auto_qmail`, being careful to name it something strange enough that no admin would happen to already have it set? Other options?

DerDakon:
I wonder why all tests pass if this should detect the regression?

schmonz:
I've made the failing test fail the build, so the checks should look uniformly bad now :-)

In an attempt to avoid

- hand-duplicating an entire rule just so I can hand-change a tiny part of it
- relying on non-portable linker tricks
- introducing a test-motivated environment variable (and function lookup) into production code
- using `chroot` or another workaround that would require running tests as root

... tests can depend on `testable`. `integrationtest_inject` depends on it, causing the top-level build to be copied into a subdirectory of `tests` where it can be messed with in unsurprising ways.

Copying seems wasteful, but for our autobuilds the time consumed is in the noise, and for local development we don't need to be running the `integrationtest` target most of the time, just the `unittest` target.

This approach seems weird but also obvious enough that we'll be unlikely to screw things up.

schmonz:
Also, the filter no longer drops a sentinel-file turd to indicate the bug. Instead it uses the customerror API from #195. Perfect fit :-)

tests/Makefile
```diff
@@ -62,3 +71,22 @@ unittest_prioq.o: \
 ../compile unittest_prioq.c
 	../compile unittest_prioq.c -I.. \
 	`pkg-config --cflags check`
+
+testable:
+	mkdir -p testable
+	cd ..; for i in *; do [ -f "$$i" ] && cp $$i tests/testable; done
```
DerDakon, line 77: This doesn't seem right. If you create ```testable``` from here, then you want to copy the things to ```testable``` as well, not ```tests/testable```. Hint: if you want to copy to a directory, then always write ```testable/```, this would bail out if the target given is not a dir and would not copy all files over each other to the same name.

gplv2/qmail-qfilter.c
```diff
@@ -0,0 +1,347 @@
...
+
+#define MSGIN 0
+#define MSGOUT 1
+#define ENVIN 3
+#define ENVOUT 4
+#define QQFD 5
+
+static const char* qqargv[2];
+
+/* a replacement for setenv(3) for systems that don't have one */
```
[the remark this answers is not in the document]
josuah: The POSIX one also has no `vallen`, I see in the rest of the code why this was desired.
This also means using memcpy for '\0'-terminated strings, which could be error-prone, but it looks correctly handled in all invocations of mysetenv.

gplv2/qmail-qfilter.c
```diff
@@ -0,0 +1,347 @@
...
+  const char* env;
+  size_t offset;
+  if ((env = mmap(0, env_len, PROT_READ, MAP_PRIVATE, ENVIN, 0)) == MAP_FAILED)
+    exit(QQ_OOM);
+  offset = parse_sender(env);
+  parse_rcpts(env, offset);
+  munmap((void*)env, env_len);
+}
+
+/* Create a temporary invisible file opened for read/write */
```
josuah, line 155: I like it! Disk storage, no cleanup needed.

josuah (approved):
This branch does not touch to any existing file beyond the Makefile.
This means very easy to integrate into an existing QMAIL setup.
It also should be straightforward to resolve the merge conflicts thanks to this.

PR #233, opened 2022-02-05
basic SMTP recipient checking

schmonz:
Give mail admins some basic SMTP recipient-checking tools:

1. An API for external programs to reject recipients (@jaysoffian's RCPTCHECK patch)
2. A program that calls the API _and_ provides the API, giving a way to list a sequence of recipient-checkers in `control/rcptchecks` (my `qmail-rcptcheck` program)
3. One such recipient-checker that requires no configuration of its own (my `qmail-rcptcheck-realrcptto`, derived from Paul Jarc's qmail-realrcptto patch)

Notably not included:

- Any other recipient-checkers, such as `qmail-rcptcheck-badrcptto` (the badrcptto patch is GPL'd, I've emailed the author) or `qmail-rcptcheck-qregex` (never needed it myself) or something for vpopmail users
- Enabling any of this by default (we probably should, in some future release)
- Adapting RCPTCHECK to `qmail-qmtpd` (probably straightforward, just needs someone who cares)
- Automated tests (I've run the code in production for years; still, sorry about this, and happy to add tests later)
- All the other programmability of [qmail-spp](http://qmail-spp.sourceforge.net) (but `qmail-rcptcheck` will happily run in that environment and recipient-checkers won't need to change at all)

### DIFF
--- a/qmail-rcptcheck-realrcptto.c
+++ b/qmail-rcptcheck-realrcptto.c
@@ -1,0 +1,27 @@
+#include <unistd.h>
+#include "env.h"
+#include "realrcptto.h"
+
+void accept_recipient() { _exit(  0); }
+void reject_recipient() { _exit(100); }
+void unable_to_verify() { _exit(111); }
+
+void die_cdb()     { unable_to_verify(); }
+void die_control() { unable_to_verify(); }
+void die_nomem()   { unable_to_verify(); }
+void die_sys()     { unable_to_verify(); }
+
+int main(void)
+{
+  char *recipient = env_get("RECIPIENT");
+
+  if (!recipient)
+    unable_to_verify();
+
+  realrcptto_init();
+  realrcptto_start();
+  if (!realrcptto(recipient))
+    reject_recipient();
+
+  accept_recipient();
+}
--- a/qmail-rcptcheck.c
+++ b/qmail-rcptcheck.c
@@ -1,0 +1,167 @@
+#include <unistd.h>
+#include "control.h"
+#include "env.h"
+#include "stralloc.h"
+#include "substdio.h"
+#include "wait.h"
+
+static char outbuf[128];
+static struct substdio ssout = SUBSTDIO_FDBUF(write,1,outbuf,sizeof outbuf);
+static char errbuf[128];
+static struct substdio sserr = SUBSTDIO_FDBUF(write,2,errbuf,sizeof errbuf);
+
+struct rcptcheck_api {
+  char *(*get_sender)();
+  char *(*get_recipient)();
+  void (*accept_recipient)();
+  void (*reject_recipient)();
+  void (*unable_to_verify)();
+  void (*unable_to_execute)();
+} rcptcheck;
+
+static char *rcptcheck_get_sender() {
+  return env_get("SENDER");
+}
+
+static char *rcptcheck_get_recipient() {
+  return env_get("RECIPIENT");
+}
+
+static void rcptcheck_accept_recipient() {
+  _exit(0);
+}
+
+static void rcptcheck_reject_recipient() {
+  _exit(100);
+}
+
+static void rcptcheck_unable_to_verify() {
+  _exit(111);
+}
+
+static void rcptcheck_unable_to_execute() {
+  _exit(120);
+}
+
+static void run_rcptchecks_under_rcptcheck() {
+  rcptcheck.get_sender        = rcptcheck_get_sender;
+  rcptcheck.get_recipient     = rcptcheck_get_recipient;
+  rcptcheck.accept_recipient  = rcptcheck_accept_recipient;
+  rcptcheck.reject_recipient  = rcptcheck_reject_recipient;
+  rcptcheck.unable_to_verify  = rcptcheck_unable_to_verify;
+  rcptcheck.unable_to_execute = rcptcheck_unable_to_execute;
+}
+
+static void putsdie(substdio *ss,char *string) {
+  substdio_puts(ss,string);
+  substdio_flush(ss);
+  _exit(0);
+}
+
+static char *spp_get_sender() {
+  return env_get("SMTPMAILFROM");
+}
+
+static char *spp_get_recipient() {
+  return env_get("SMTPRCPTTO");
+}
+
+static void spp_accept_recipient() {
+  putsdie(&ssout,"");
+}
+
+static void spp_reject_recipient() {
+  putsdie(&ssout,"E553 sorry, no mailbox here by that name. (#5.1.1)\n");
+}
+
+static void spp_unable_to_verify() {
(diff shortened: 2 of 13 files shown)


realrcptto.c
```diff
@@ -0,0 +1,346 @@
...
+ * TODO:
+ * "copies logic from qmail-send, qmail-lspawn, qmail-getpw, and qmail-local"
+ * Reduce this duplication: factor out what we need from these programs,
+ * then link with those objects at their original sites and also here.
+ */
+
+extern void die_nomem();
+extern void die_control();
+extern void die_cdb();
+extern void die_sys();
```
DerDakon, line 31: Why aren't they in realrcptto.h? They also need ```_noreturn``` annotations.

qmail-rcptcheck-realrcptto.c
```diff
@@ -0,0 +1,27 @@
+#include <unistd.h>
+#include "env.h"
+#include "realrcptto.h"
+
+void accept_recipient() { _exit(  0); }
+void reject_recipient() { _exit(100); }
+void unable_to_verify() { _exit(111); }
```
DerDakon, line 7: They should be annotated ```__noreturn```.

qmail-rcptcheck.c
```diff
@@ -0,0 +1,167 @@
...
+#include "wait.h"
+
+static char outbuf[128];
+static struct substdio ssout = SUBSTDIO_FDBUF(write,1,outbuf,sizeof outbuf);
+static char errbuf[128];
+static struct substdio sserr = SUBSTDIO_FDBUF(write,2,errbuf,sizeof errbuf);
+
+struct rcptcheck_api {
+  char *(*get_sender)();
+  char *(*get_recipient)();
```
DerDakon, line 15: The return type should be const to make it clear that the caller may not modify the value.

qmail-rcptcheck.c
```diff
@@ -0,0 +1,167 @@
...
+  switch(wait_exitcode(wstat)) {
+    case 100: rcptcheck.reject_recipient();
+    case 111: rcptcheck.unable_to_verify();
+    case 120: rcptcheck.unable_to_execute();
+    default:  return;
+  }
+}
+
+static int looks_like_spp() {
+  char *x = env_get("SMTPRCPTTO");
```
DerDakon, line 122: Should be const as well.

qmail-rcptcheck.c
```diff
@@ -0,0 +1,167 @@
...
+
+static void run_rcptcheck(char *program)
+{
+  char *rcptcheck_program[2] = { program, 0 };
+  int pid;
+  int wstat;
+
+  switch(pid = fork()) {
+    case -1:
+      rcptcheck.unable_to_execute();
```
DerDakon, line 102: The callbacks are expected to be noreturn here, otherwise one would fallthrough. Without annotations this will lead to compiler warnings.

I wonder if it isn't less error prone to expect the callbacks to return and instead exit in the caller. That way an unexpected return from such a callback will not screw up the program logic.

DerDakon (changes requested):

qmail-rcptcheck.c
```diff
@@ -0,0 +1,167 @@
...
+
+int main(void)
+{
+  stralloc rcptchecks = {0};
+  int linestart;
+  int pos;
+
+  if (looks_like_spp()) run_rcptchecks_under_spp();
+  else run_rcptchecks_under_rcptcheck();
+
```
josuah, line 153: At this point, we should not have any empty function pointer.
Checking... Each `run_rcptchecks_under_*` sets 6 function pointers, the struct has 6 function pointers, run in an if-else... Function pointers not called before... Looks all good.

badrcptto.c
```diff
@@ -0,0 +1,67 @@
...
+}
+
+static int _badrcptto_reject_whole_domain(struct constmap map, stralloc address)
+{
+  /* why not just comment out the domain in control/rcpthosts? */
+  int j = byte_rchr(address.s,address.len,'@');
+  return ((j < address.len) && (constmap(&map,address.s + j,address.len - j - 1)));
+}
+
+static int _badrcptto_reject_string(char *string)
```
josuah, line 35: A `char *` string passed without a length looks suspicious: would `char *string` be written to without a length? Of course not, it is only read.

Making it `char const *` saves the time it takes to check it, and make it more obvious that the function returns 0 or 1 depending on `string`.

badrcptto.c
```diff
@@ -0,0 +1,67 @@
...
+
+extern void die_control();
+extern void die_nomem();
+
+static void _badrcptto_log_rejection(char *recipient)
+{
+  char smtpdpid[32];
+  char *remoteip = env_get("TCPREMOTEIP");
+  if (!remoteip) remoteip = "unknown";
+  str_copy(smtpdpid + fmt_ulong(smtpdpid,getppid())," ");
```
josuah, line 19: Same question with str_copy vs strcpy.

badrcptto.c
```diff
@@ -0,0 +1,67 @@
...
+{
+  /* why not just comment out the domain in control/rcpthosts? */
+  int j = byte_rchr(address.s,address.len,'@');
+  return ((j < address.len) && (constmap(&map,address.s + j,address.len - j - 1)));
+}
+
+static int _badrcptto_reject_string(char *string)
+{
+  stralloc addr = {0};
+  stralloc brt = {0};
```
josuah, line 38: I think I missed when those two are set `stralloc_free()`. Or does it not matter since it is a short-lived program?

I realize that there is not even any [stralloc_free()](https://skarnet.org/software/skalibs/libstddjb/stralloc.html) in DJB code. It could be introduced later.

josuah: > The benefits of using stralloc are as follows: [...]
> - If every heap-allocated object is represented by a stralloc, then it is very easy to identify what pointer is in the heap. When you stop using a pointer p, should you free it ? Sometimes it's not easy to find out. When you stop using a stralloc sa, you know you must call stralloc_free(&sa). Store your strong references as strallocs and weak references as simple pointers, and never free simple pointers. This policy allows me to boast that no skarnet.org software has ever leaked memory.  -- https://skarnet.org/software/skalibs/libstddjb/stralloc.html

josuah: Other wraparound of DJB code do include `stralloc_free`, like fefe's https://www.fefe.de/libowfat/:

```
#include <stdlib.h>
#include "stralloc.h"

void stralloc_free(stralloc *sa) {
  if (sa->s) free(sa->s);
  sa->s=0;
  sa->a=sa->len=0;
}
```

josuah (changes requested):
That is a great addition, enters the game of providing an API, and a simple and sane one, for making patches unneeded. Kudos for bringing that up.

Letting you adjust the `__noreturn` and `char` -> `char const` as proposed by @DerDakon.

I will open an issue for the other points I discussed.

Thank you!

PR #234, opened 2022-05-03
qmail-qstat: simplify the code

DerDakon:
See the commits for a detailed description, but boils down to:

* avoid needless dependencies in Makefile
* use ```find -type f``` to just count files and avoid much complication in the script

### DIFF
--- a/qmail-qstat.sh
+++ b/qmail-qstat.sh
@@ -1,7 +1,4 @@
-cd QMAIL
-messdirs=`echo queue/mess/* | wc -w`
-messfiles=`find queue/mess/* -print | wc -w`
-tododirs=`echo queue/todo | wc -w`
-todofiles=`find queue/todo -print | wc -w`
-echo messages in queue: `expr $messfiles - $messdirs`
-echo messages in queue but not yet preprocessed: `expr $todofiles - $tododirs`
+messfiles=`find QMAIL/queue/mess/* -type f | wc -l`
+todofiles=`find QMAIL/queue/todo -type f | wc -l`
+echo "messages in queue:" $messfiles
+echo "messages in queue but not yet preprocessed:" $todofiles
--- a/Makefile
+++ b/Makefile
@@ -1324,11 +1324,9 @@ exit.h
 	./compile qmail-qread.c
 
 qmail-qstat: \
-warn-auto.sh qmail-qstat.sh conf-qmail conf-break conf-split
+warn-auto.sh qmail-qstat.sh conf-qmail
 	cat warn-auto.sh qmail-qstat.sh \
 	| sed s}QMAIL}"`head -n 1 conf-qmail`"}g \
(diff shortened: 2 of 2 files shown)


DerDakon:
@mbhangui your Indimail uses big-todo IIRC, could you check if this works properly there?

mbhangui (approved):

mbhangui:
> @mbhangui your Indimail uses big-todo IIRC, could you check if this works properly there?

This works for big-todo because it uses find. The earlier script didn't work because it expected all files to be in queue/todo directory.

qmail-qstat.sh
```diff
@@ -1,7 +1,4 @@
-cd QMAIL
-messdirs=`echo queue/mess/* | wc -w`
-messfiles=`find queue/mess/* -print | wc -w`
-tododirs=`echo queue/todo | wc -w`
-todofiles=`find queue/todo -print | wc -w`
-echo messages in queue: `expr $messfiles - $messdirs`
-echo messages in queue but not yet preprocessed: `expr $todofiles - $tododirs`
+messfiles=`find QMAIL/queue/mess/* -type f | wc -l`
+todofiles=`find QMAIL/queue/todo -type f | wc -l`
```
josuah, line 2: The `wc -w` to `wc -l` make sense, find produces one result per line.
It is all based upon the absence of spaces in file names here anyway.

Maybe another case where `-type f` was not widely available.

josuah (approved):
LGTM, seal of approval

Maybe if someone does something funky with symlinks, then an extra `-L` flag to find could be useful though?

qmail-qstat.sh
DerDakon: I expect someone that does such special things to the queue to be able to hack this script as well.

schmonz:
`find(1)` is always a portability risk, but the way it's used here feels likely to work on every platform we could possibly care about, and the way this would break (if we're wrong about that) would be tolerable. Breaking this portion of the big-todo patch would usually be counter to our project goals, but I consider this an excellent exception: we're breaking it because we've made it unnecessary, and the rest of the patch is unaffected. And of course removing unused dependencies is a small but permanently cost-cutting victory.

qmail-qstat.sh
```diff
@@ -1,7 +1,4 @@
...
+echo "messages in queue: $messfiles"
+echo "messages in queue but not yet preprocessed: $todofiles"
```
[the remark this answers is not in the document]
josuah: The possible reason:

```
$ echo "$(echo "   :")"
   :
```

Workaround 1: adding `| tr -d '\t '` after find

Workaround 2: remove the quotes after echo:

```suggestion
echo messages in queue: $messfiles
echo messages in queue but not yet preprocessed: $todofiles
```

schmonz (approved):

PR #237, opened 2022-07-04
add missing headers for some system functions

DerDakon:
This has been split out of #65. The commit about headers for ```umask()``` has been omitted as ```mode_t``` is not used, so ```<sys/types.h>``` isn't actually needed anywhere for this.

### DIFF
--- a/qmail-remote.c
+++ b/qmail-remote.c
@@ -2,6 +2,7 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <unistd.h>
 #include "sig.h"
 #include "stralloc.h"
 #include "substdio.h"
--- a/control.c
+++ b/control.c
@@ -1,5 +1,6 @@
 #include "control.h"
 
+#include <unistd.h>
 #include "readwrite.h"
 #include "open.h"
 #include "getln.h"
--- a/instcheck.c
+++ b/instcheck.c
@@ -1,6 +1,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <string.h>
+#include <unistd.h>
 #include "strerr.h"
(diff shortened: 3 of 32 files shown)


mbhangui (approved):
LGTM

leahneukirchen (approved):

schmonz:
Is there a particular compiler warning level you're using to make these pop out? Something we could include in one of the autobuilds?

qmail-remote.c
```diff
@@ -2,6 +2,7 @@
 #include <sys/socket.h>
 #include <netinet/in.h>
 #include <arpa/inet.h>
+#include <unistd.h>
```
schmonz, line 5: In general this change is of course correct and needed, and as usual I'm mildly worried only about the timing of the change in programs which many users are likely to still be heavily patching: this one and `smtpd`, mainly.

@schmonz gonna schmonz ;-) (and seriously, it's nice to be back here thinking about notqmail, so thank you) -- but I would _love_ to be able to feel that this sort of thing is getting worried about automatically. Then I could find new kinds of things to be mildly annoying about. @josuah, how can we help finish getting the patch-merge-checker into our regular autobuilds?

schmonz (approved):
In a different situation I might want to block this on having our autobuilds auto-detect popular merge conflicts. Right now it's more important to me that we start moving again. Thank you for the impetus.

qmail-remote.c
DerDakon: As this a bit older: no, I don't think so. At least I can't provoke a warning on my system, most likely because (at least today, maybe not when these patches were originally done) these headers are in one or another indirect way automatically included. So for now I would call it "manual code audit", even if it was not true back then it would be true if I had to do them again.

PR #240, opened 2022-10-13
CI: update GitHub actions to Ubuntu 22.04

DerDakon:
This already includes a newer version of check, and the mirror used no longer has hirsute available, so CI builds are currently failing.

While at it bump the version of the checkout action.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        host: [ubuntu-latest, macos-latest]
+        host: [ubuntu-22.04, macos-latest]
         config:
          - {
            name: gcc-default,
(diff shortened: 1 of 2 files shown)


mbhangui (approved):
LGTM

schmonz (approved):
Sorry for the hard-to-guess failure, and thanks for looking after it. Next time I do something like this, I’ll try to remember to include an OS version check so we’ll be given a better clue when it’s time to change.

leahneukirchen (approved):

josuah (approved):
Looks harmless to me.

The fact that we stick to one particular Ubuntu version could help to differentiate:
* a new version of ubuntu breaking something
* a commit pushed breaking something

DerDakon:
The reason I use "22.04" instead of "latest" is just that the latter is currently still an alias for "20.04". But I think doing an explicit upgrade here is no bad step, as we can do a check (sic!) before the switch to make sure it actually works.

PR #242, opened 2022-10-21
CI: update Solaris and add OpenBSD runner

DerDakon:


.github/workflows/openbsd.yml
```diff
@@ -0,0 +1,48 @@
...
+      matrix:
+        config:
+         - {
+           name: default64,
+           nroff: nroff,
+           cflags: ""
+         }
+         - {
+           name: no_obsolete64,
+           nroff: true,
```
leahneukirchen, line 28: I think this is yaml and needs quoting here, else it's a bool

schmonz: If that's the case, we need it fixed across all the GitHub builders.

DerDakon: At least it works, because the build runs "true -man something.8".

leahneukirchen: It probably works because the stringification turns it into "true".

Good thing we are not testing "yes" instead. ;)

mbhangui (approved):

.github/workflows/openbsd.yml
DerDakon: Works quoted as well, so I quote it in the new file.

.github/workflows/openbsd.yml
```diff
@@ -0,0 +1,48 @@
...
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: OpenBSD-${{ matrix.config.name }}
+    runs-on: macos-12
```
josuah, line 13: `macos-12` is accurate, but confusing. What about a comment above this line? Eventually a link to https://github.com/marketplace/actions/openbsd-vm

.github/workflows/openbsd.yml
```diff
@@ -0,0 +1,48 @@
...
+    env:
+      CFLAGS: ${{ matrix.config.cflags }}
+      NROFF: ${{ matrix.config.nroff }}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+         - {
+           name: "default64",
+           nroff: "nroff",
```
josuah, line 23: In case this fails, maybe `mandoc` would work better then.

josuah (approved):
All good! Thank you for pushing the work of importing these.

Testing on so many different platforms will help with removing the "compat workarounds" through the source.

schmonz (approved):

.github/workflows/openbsd.yml
DerDakon: Done.

.github/workflows/openbsd.yml
DerDakon: At the moment it just works.

PR #245, opened 2023-10-19
1.09 release commits

schmonz:


mbhangui (approved):

DerDakon (approved):

DerDakon:
Needs https://github.com/notqmail/notqmail/pull/246 to fix the build first.

josuah (approved):
Thank you all, for this amazing continued effort!
This CHANGES.md goes back all the way up to 1996, and still around today!

josuah:
Good idea to also document the CI changes, this can be insightful for someone interested in notqmail.

leahneukirchen (approved):

DerDakon:
> Good idea to also document the CI changes, this can be insightful for someone interested in notqmail.

I would actually see it the other way 'round: a user of the release tarball does not care for that, as those files are not shipped with the tarballs anyway.

PR #246, opened 2023-10-19
tests: fix return type read() and write() replacement stubs

DerDakon:
Detected by build breakage with FreeBSD 14 CI.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -67,7 +67,7 @@ jobs:
 
     steps:
     - name: checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: test_deps
       run: |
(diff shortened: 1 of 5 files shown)


josuah:
I agree that ssize_t sounds suitable here, though I am confused about a `size_t` variable returned as `ssize_t`: it looks like either inlen/outlen should be turned into `ssize_t` as well, either the function may return `size_t`.

mbhangui (approved):

DerDakon:
> I agree that ssize_t sounds suitable here, though I am confused about a `size_t` variable returned as `ssize_t`: it looks like either inlen/outlen should be turned into `ssize_t` as well, either the function may return `size_t`.

The interface must match that one of `read()` and `write()`, and if I actually did that correctly the last argument would have been a `size_t` as well (which it is now), and then it makes sense to compare against a local size_t as well. These functions are returning wrapped around values on some platforms if you write e.g. 3GB in a single call on 32 bit. It's then only the -1 that is actually special.

schmonz (approved):

josuah (approved):
Thank you for your clear explanations. I understand the rationale now. :)

PR #253, opened 2024-01-13
Repair autobuilds

schmonz:
[vmactions](https://github.com/vmactions) has new versions that work differently under the hood: instead of relying on VirtualBox on a particular old macOS, they run `qemu` under `ubuntu-latest`. These seem as quick as they used to be (including Solaris, which is back to needing 10 minutes or so), hopefully reliable again, and almost certainly less likely to break over time.

With FreeBSD the only one not on Actions (CirrusCI) and the only one still not working, I converted it to Actions as well.

I _think_ I had been seeing each of our jobs running twice on every push to a branch that's part of a PR: once on `push`, and once on `pull_request`. If I misinterpreted what I saw, then the `on:` stanzas (and the `if: github.event_name == 'push' || github.event.pull_request.merged == true)`) are unnecessarily complicated and we can put them back as they were. Also, I haven't tested that the jobs really do run when a PR is merged! When we merge this one, we have to watch carefully.

Now that feedback is fast again, I've expanded scope, adding NetBSD (`x86_64`) and Alpine (many architectures). Alpine's `s390x` build shows a bug in one of our tests!

Now that there are many more variants, I've changed how jobs are named: "defaults" (e.g., `gcc`, 64-bit ABI, `nroff`) are left out, only "non-defaults" (e.g., `clang`, 32-bit ABI, `mandoc` or `true`) are shown. I'm hoping this makes it easier to differentiate when something's broken.

Some of the variants are probably redundant. For example, the `nroff` and `no-roff` variants are probably covered well enough with one BSD, one GNU, and one Solaris `make` (and then the platforms with `mandoc` in place of `nroff` can specify no more than that). Likewise, `-DDEPRECATED_FUNCTIONS_REMOVED` probably needn't be a variant in every single build.

FreeBSD does a whole `make package` run-through, with and without `DESTDIR`. FreeBSD also has possibly the most thorough `-W` settings, to make sure we don't backslide on classes of warnings we've fixed. Maybe we want to extend one or both of these to all platform builders?

Whenever it was that the VM builders suddenly got unreliable, I felt frustrated and a bit powerless. Of course it would have been smart for me to check whether there were new @vmactions ready (or being worked on), but I didn't think of that until very recently. So I'm wishing for an automated way to be notified when @vmactions updates come out and/or the available OS version list changes, or the CodeQL action, or the `ubuntu` or `macos` version lists.

There's a lot of duplication across the `.yml` files. I don't think this PR should wait until we figure out how to reduce the duplication, but I'm hoping it'll get better as we learn more about Actions.

Solaris might be a little faster with [the cache action](https://github.com/actions/cache) reusing our hand-built `libcheck` artifact, but will probably keep being slow no matter what.

Once this is ready to merge, I'll make sure the `master` branch protection rules are updated to require all the variants we decided to keep.

My next assignment should probably be #225!


mbhangui (approved):

DerDakon:
FreeBSD builds had a 3x2 matrix, i.e. all 3 versions used both nroff and true. Did you change that intentionally?

schmonz:
> FreeBSD builds had a 3x2 matrix, i.e. all 3 versions used both nroff and true. Did you change that intentionally?

Sorta -- I figured we had enough coverage of those combinations in the OpenBSD builds. But given how quick these builds seem to be, happy to put it back.

josuah (approved):
I was wondering where all these vmactions were coming from...
https://github.com/vmactions/vbox/blob/main/vbox.sh

It is all qemu under the hood! Big joke on Github to let users spin VM to have whichever system they actually want!

I suppose Github will come with a home-grown solution to avoid users spending so many minutes at some point... But this is Github problem to solve, and it makes sense to just jump in the train until then!

Thank you once again for this continuous effort for maintainance. On my side, I am under water with firmware development, which I do exclusively these days.

.github/workflows/native-ubuntu.yml
```diff
@@ -10,35 +10,38 @@ on:
...
       fail-fast: false
       matrix:
         osversion: ["22.04"] # 20.04's "check" package is too old
         abi: [64]
         cflags: ["-DDEPRECATED_FUNCTIONS_REMOVED -Werror=deprecated-declarations"] # ""
+        warnings: "-Wall -Wshadow -Werror=implicit-function-declaration -Werror=parentheses -Werror=dangling-else -Werror=pointer-sign"
         cc: ["gcc", "clang"]
-        nroff: ["nroff", "true"]
-        utmpflags: ["", "-o qtmp.h"]
+        nroff: ["nroff", "no-roff"]
```
josuah: Less confusing this way, I agree.

.github/workflows/native-macos.yml
```diff
@@ -10,34 +10,36 @@ on:
...
     steps:
     - name: Checkout
       uses: actions/checkout@v4
-    - name: Test
-      id: test
-      env:
-        CC: ${{ matrix.cc }}
-        CFLAGS: "${{ matrix.cflags }} -O2 -fPIC -Wall -Wshadow -Werror=implicit-function-declaration ${{ matrix.cc == 'clang' && '-Werror=incompatible-library-redeclaration' || ' ' }} -Werror=parentheses -Werror=dangling-else -Werror=pointer-sign -Werror=pointer-sign"
-        NROFF: ${{ matrix.nroff }}
-        UTMP: ${{ matrix.utmp }}
```
josuah: Always nice to have fewer boilerplate. The lines end-up slightly longer, but still readable.

.github/workflows/vm-alpine.yml
```diff
@@ -30,13 +34,11 @@ jobs:
...
         arch: ${{ matrix.platform }}
         packages: "build-base check check-dev pkgconf"
-    - name: Build
+    - name: Build and Test
       shell: alpine.sh {0}
       run: |
+        set -e
         echo "cc -O2 ${{ matrix.cflags }}" > conf-cc
         echo "cc -s" > conf-ld
         make -j 2 it man NROFF=${{ matrix.nroff }}
```
josuah: Good catch! Then if `make man` fails, we are warned about it.

.github/workflows/vm-freebsd.yml
```diff
@@ -10,13 +10,14 @@ on:
 
 jobs:
   build-freebsd:
-    if: false && (github.event_name == 'push' || github.event.pull_request.merged == true)
-    name: FreeBSD ${{ matrix.osversion }} ${{ matrix.abi == '64' && ' ' || matrix.abi }} ${{ matrix.nroff == 'true' && 'no-roff' || ' ' }} ${{ matrix.cflags == '' && ' ' || 'max-Werror' }}
+    if: true && (github.event_name == 'push' || github.event.pull_request.merged == true)
```
josuah: If I understand, this "false" -> "true" is there to quickly toggle the FreeBSD build on/off?
Having `true && expression` is a bit surprising, but this is also infrastructure configuration meant for maintainers, so good to have a quick toggle

.github/workflows/vm-netbsd.yml
```diff
@@ -10,34 +10,35 @@ on:
...
-    env:
-      CFLAGS: ${{ matrix.cflags }}
-      NQABI: ${{ matrix.abi }}
-      NROFF: ${{ matrix.nroff }}
     strategy:
       fail-fast: false
       matrix:
         osversion: ["9.3"]
-        abi: [64] # 32
+        abi: [64, 32]
```
josuah: This means you are trying 32-bit binaries on a 64-bit architecture? If so, quite extensive tests!

josuah (approved):
That is a lot fewer declarations this way.
We will quickly see if this still works as intended upon the next runs.

Thank you for this refresh-up!

.github/workflows/native-macos.yml
schmonz: Yes, my feeling is that the the tradeoff is desirable: the job gets more glanceable than the line lengths get unreadable ;-)

.github/workflows/vm-alpine.yml
schmonz: Thanks to @lighttpd (I saw this in their [pr.yml](https://github.com/lighttpd/lighttpd1.4/blob/79d011de6f942b2607f4d012e0372fb966c14bcf/.github/workflows/pr.yml#L35)).

.github/workflows/vm-freebsd.yml
schmonz: Yes, exactly, this was to to speed up the GitHub Actions development cycle a bit. Not sure it's worth keeping around. If there were a more declarative way to temporarily skip some...

.github/workflows/vm-netbsd.yml
schmonz: Not currently working (because 32-bit tests won't link with 64-bit `libcheck`) and seems not worth the trouble, so I'm gonna remove it. I _would_ like a variety of NetBSD hardware architectures, as we now have with Alpine, but am not aware of any Actions designed for that. (My [nbvm](https://github.com/schmonz/nbvm) project may help someday but is nowhere near ready.)

.github/workflows/platform-builders.yml
```diff
@@ -0,0 +1,264 @@
...
+    branches: ['**', '!master']
+  pull_request_target:
+    branches: ['master']
+    types:
+      - closed
+
+jobs:
+
+  alpine:
+    if: github.event_name == 'push' || github.event.pull_request.merged == true
```
DerDakon, line 14: Is this still needed? The build should only run at this time based on the ```on:``` settings, no?

DerDakon:
I would like to see that the build indeed runs at merge as intended, otherwise LGTM.

.github/workflows/platform-builders.yml
schmonz: What I'm trying for here is a combination of
1. Trigger the workflow when we push to any non-`master` branch, whether it's affiliated with a PR or not
2. Avoid triggering the same workflow a second time in the (common) case where there _is_ a PR for the branch
3. Trigger the workflow when the PR is merged to `master`

The `'push'` part of the conditional indeed seems redundant with what's in `on: push`. If the second trigger could be for `on: pull_request_target: types: merged`, then I think we could get rid of the entire `if:` line for each job. But my understanding (from [Stack Overflow](https://stackoverflow.com/questions/60710209/trigger-github-actions-only-when-pr-is-merged)) is that we can't say `merged` in the trigger, only `closed`, which means that in jobs we have to check conditionally which kind of `closed` it was. :-/

schmonz:
> I would like to see that the build indeed runs at merge as intended

https://github.com/notqmail/notqmail/actions/runs/7614561039 was triggered on `closed`. I guess we haven't tested that closing-without-merging a PR does _not_ trigger the workflow, but probably not, and I'm comfortable with that risk. ;-)

DerDakon (approved):

schmonz:
Merged, and updated the `master` branch protection rules to correspond to these builds (plus the Analyze check).

PR #254, opened 2024-01-17
tests: fix array sorting on big endian 64 bit

DerDakon:
The array elements are unsigned long, i.e. 64 bit on such arches. When casting them to int only the lower half would be compared, which doesn't matter on 32 bit arches or little endian, but fails on 64 bit BE as s390 or sparc.


mbhangui (approved):

schmonz (approved):
Very good chance this was my mistake.

josuah (approved):
Good catch! A proof that testing all sorts of endianess and architecture helps...

PR #255, opened 2024-01-19
Don't generate empty catpages by mistake.

schmonz:
We have a mechanism for not installing preformatted manpages: pass `NROFF=true` on the `make` command line, thereby generating `.0` files that are empty, which are then intentionally ignored by the installer.

We currently also generate empty `.0` files in the case where the default or overridden `NROFF` command is not found (or has a runtime error), so the `man` target appears to succeed and the resulting installation lacks the expected catpages.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,20 @@ default: it
 .SUFFIXES: .0 .1 .5 .7 .8
 
 .1.0:
-	$(NROFF) -man $< >$@
+	$(NROFF) -man $< >$@; \
+	ret=$$?; [ 0 = $$ret ] || rm -f $@; exit $$ret
 
 .5.0:
-	$(NROFF) -man $< >$@
+	$(NROFF) -man $< >$@; \
+	ret=$$?; [ 0 = $$ret ] || rm -f $@; exit $$ret
 
 .7.0:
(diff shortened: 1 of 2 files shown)


mbhangui (approved):

DerDakon (approved):

josuah (approved):
I think there is `DELETE_ON_ERROR` that does it, but that would make it less portable.
Now, if $NROFF fails, it actually ignores the error, and does not generate the `.0`, silently failing: `||` masks the exit code

Maybe the next step that called that  recipe would error though, such as the one installing the catpages... And then the error would get reported that way.

If you are fine with this scheme, then let's do this.

Otherwise, this can be modified that way:

```
if $(NROFF) -MAN $< >$@; then exit 0; else rm -f $@; exit 1; fi
```
A bit more ugly, a bit more explicit about shell `&&` and `||` trickery.

josuah:
Must I not forget to thank you!

schmonz:
Good catch, @josuah (my turn to thank you). I've updated this to propagate the exit code from `$(NROFF)`, as you suggest, in a way that feels perhaps marginally easier to read.

PR #261, opened 2024-01-31
Remove @leahneukirchen from CODEOWNERS.

schmonz:


mbhangui (approved):

josuah (approved):
I've learned plenty about email from Leah here. That is some great gift left behind.
o7

PR #264, opened 2024-02-02
autobuilds: speed up Solaris, add DragonFly

schmonz:
Potential platforms that are available via GitHub Actions, I've tested recently with another project, and will not be adding to our matrix at present:

1. OpenBSD/arm64 and FreeBSD/arm64 (7.4 and 13.2 are latest at time of writing), via @cross-platform-actions:
    - In my testing for another project, they take an order of magnitude longer to run
    - We have a nonzero number of arm64 builds already, thanks to Alpine
2. OmniOS, via @vmactions:
    - As with Solaris, there's no native package of [check](https://libcheck.github.io/check/)
    - I'm not eager to be hand-building `libcheck` in two places
    - Building on Oracle Solaris means our chances of building on Illumos are pretty good

Once arm64 OpenBSD, FreeBSD, and (hopefully soon) NetBSD are running acceptably fast under the new `macos-14` host that's aarch64, we should add them.

Once our Solaris builder is caching our hand-built `libcheck`, we can reconsider the cost/benefit of adding OmniOS.

### DIFF
--- a/.github/workflows/platform-builders.yml
+++ b/.github/workflows/platform-builders.yml
@@ -36,6 +36,37 @@ jobs:
         make -j 2 it man NROFF=${{ matrix.nroff == 'no-roff' && 'true' || matrix.nroff }}
         make -j 2 test
 
+  dragonflybsd:
+    name: >
+      DragonFlyBSD
+      ${{ matrix.osversion }}
+      ${{ matrix.abi == '64' && ' ' || matrix.abi }}
+      ${{ matrix.nroff == 'nroff' && ' ' || matrix.nroff }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        osversion: ["6.4.0"]
+        abi: [64]
+        nroff: ["mandoc", "no-roff"]
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+    - name: Build and Test
+      uses: vmactions/dragonflybsd-vm@v1
+      with:
+        release: ${{ matrix.osversion }}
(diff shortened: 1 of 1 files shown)


josuah (approved):
Yet another distribution supporting notqmail!
Does that mean it should be added to this page?
https://notqmail.org/platforms/

The automation infrastructure itself is becoming quite advanced!

schmonz:
> https://notqmail.org/platforms/

Good call, that page was outdated and definitely needed a little something. I added some explanatory text about what's covered and where we have gaps, along with a link to our autobuild results. (The old findings are still in place.)

> The automation infrastructure itself is becoming quite advanced!

I recently packaged [Leah's mlog](https://github.com/leahneukirchen/mlog) [for pkgsrc](https://github.com/schmonz/pkgsrc-mlog/), and have been [experimenting with multi-platform builds](https://github.com/schmonz/pkgsrc-mlog/actions/runs/7750326613) there and bringing whatever works well back here.

DerDakon (approved):
Nice.

What I wonder is if there is a non-gcc option available for Solaris as well?

schmonz:
> What I wonder is if there is a non-gcc option available for Solaris as well?

The vendor still offers a compiler (with the name Studio in it, IIRC), and it'd make a great addition to the build mix. A good idea in general to diversify not only OSes and architectures but also compilers. I vaguely recall having gotten Studio going someplace once, but will probably have to rediscover the necessary incantations.

schmonz:
Post-merge, added the two DragonFly builds to the protection rules for `master`.

PR #266, opened 2024-02-02
instchown: operate on fds, not filenames.

schmonz:
The previous code had been flagged by CodeQL as "TOCTOU". It's not -- at worst, it's time-of-chown to time-of-chmod -- but this will run more efficiently (and make the warnings go away).

Tested like so:

```sh
#!/bin/sh

set -e

for i in true mandoc; do
	sudo rm -rf /var/tmp/qmail-staging $(head -1 conf-qmail)
	make clean
	make -j 4 it man NROFF=$i test
	env DESTDIR=/var/tmp/qmail-staging make package
	sudo env DESTDIR=/var/tmp/qmail-staging ./instchown
	sudo make package
	sudo ./instchown
	sudo ./instcheck
done
```

### DIFF
--- a/instchown.c
+++ b/instchown.c
@@ -4,6 +4,7 @@
 #include <string.h>
 #include <unistd.h>
 #include "env.h"
+#include "open.h"
 #include "stralloc.h"
 #include "strerr.h"
 #include "error.h"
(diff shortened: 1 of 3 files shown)


mbhangui (approved):

schmonz:
Yes, AFAICT it's limited to time-of-chown vs time-of-chmod. But (if I can get it to actually work) as you say, seems wise to not be in position to find out whether it's anything worse.

instchown.c
```diff
@@ -81,28 +101,39 @@ void c(char *home, char *subdir, char *file, uid_t uid, gid_t gid, int mode)
...
   }
-  if (chown(file,uid,gid) == -1) {
-    /* assume cat man pages are simply not installed */
-    if (errno == error_noent && strncmp(subdir, "man/cat", 7) == 0)
-      return;
-    strerr_die6sys(111,FATAL,"unable to chown .../",subdir,"/",file,": ");
+  if ((fd = open_read(file))) {
+    if (fchown(fd,uid,gid) == -1) {
+      /* assume cat man pages are simply not installed */
+      if ((errno == error_noent || errno == EBADF) && strncmp(subdir, "man/cat", 7) == 0)
```
DerDakon, line 107: use ENOENT directly

instchown.c
```diff
@@ -30,46 +31,65 @@ static void ddhome(stralloc *dd, const char *home)
...
 {
+  int fd;
   stralloc dh = { 0 };
   ddhome(&dh, home);
   home=dh.s;
-  if (chown(home,uid,gid) == -1)
-    strerr_die4sys(111,FATAL,"unable to chown ",home,": ");
-  if (chmod(home,mode) == -1)
-    strerr_die4sys(111,FATAL,"unable to chmod ",home,": ");
+  if ((fd = open_read(home))) {
```
DerDakon, line 38: This will not work at all because on error this returns -1 which is "true". You need ```((fd = open_read(home)) >= 0)```.

DerDakon (commented):
While I think the automated report it rather pointless and this is no real issue, I actually like that change because it is way more efficient for the OS at the end.

DerDakon (changes requested):
wrong checkbox

DerDakon (approved):

PR #267, opened 2024-02-06
add queue-fix for queue creation/repair

mbhangui:
New program queue-fix which can create or repair a queue directory passed to it on the command line

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,6 @@
+- 20240206 code: add queue-fix to create qmail queue from scratch,
+           fix existing queue and set proper permissions for all directories,
+           files, fifo which are part of the queue
 - 20240127 remove instqueue, now replaced by "instpackage queue-only"
 - 20221115 honor all groups of qmailr user when spawning qmail-rspawn
 - 20221019 code: netstring parsing in qmail-qmtpd & qmail-qmqpd now
--- a/Makefile
+++ b/Makefile
@@ -538,6 +538,10 @@ fifo.o: \
 compile fifo.c hasmkffo.h fifo.h
 	./compile fifo.c
 
+fmt_8long.o: \
+compile fmt_8long.c fmt.h
+	./compile fmt_8long.c
+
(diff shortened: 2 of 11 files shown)


schmonz:
Last time we thought about this together, which admittedly was quite some time ago, we thought a sysadmin-facing queue fix/create tool [would belong in 2.0](https://notqmail.org/roadmap) (or maybe slightly before that, to aid in the transition of the queue format).

We do usually like to borrow well-loved code that qmail users have been relying on, but I'm not sure we've put in enough consideration yet for what we'll need our own queue-maintenance tool to be able to do and how best to have it do that.

Should we be considering this PR with urgency for 1.09? as a helpful stopgap soonish?

I like `queue-fix` and appreciate what you've done with it, just trying to understand where we want this to fit in our plans.

mbhangui:
> 
> We do usually like to borrow well-loved code that qmail users have been relying on, but I'm not sure we've put in enough consideration yet for what we'll need our own queue-maintenance tool to be able to do and how best to have it do that.
> 

The most important feature of queue-fix isn't the ability to create a queue during installation, but repair a damaged queue during a filesystem corruption. This situation will not likely happen with normal user. But very likely to happen on a mail server with heavy traffic where the queue filesystem is the most heavily used. Hard reboot of such a server has the potential of damaging the filesystem on which /var/qmail/queue lies. When this happens it is sometime impossible to queue new messages. The qmail-send logs too get filled with errors. queue-fix is a quick way to fix the queue if the server doesn't have the source code.

> Should we be considering this PR with urgency for 1.09? as a helpful stopgap soonish?
> 

I feel there is no urgency. We have instpackage queue-only option to create the queue. Most notqmail users are adept at installing patches and can always get the queue-fix program from various sources. But having queue-fix is more of a long term strategy where there isn't a single tool missing in the notqmail arsenal that helps the end user fix issues.

josuah:
That sounds also a lot alike notqmail to try to integrate the features that are the most expected by their users.

PR #271, opened 2024-03-23
clean up interface of substdio functions

DerDakon:
Get rid of declaration without arguments in the header, as that causes warnings in most recent compilers. While at it clean up a bit more.

I'm not aware of any patch that modifies substdio.h or any of it's functions, so I think this is safe.


josuah (approved):
Woops! Yes, much needed changes...
The `inline` are also a good thing to have, in particular when `strlen()` is involved.

schmonz (approved):

PR #272, opened 2024-03-23
always use the passed fd in generated saferead() and safewrite() functions

### LINKED ISSUE #270: Why do saferead and safewrite sometimes ignore the file descriptor that is passed to them?
The new ```GEN_SAFE_TIMEOUTREAD``` and ```GEN_SAFE_TIMEOUTWRITE``` makros allow to pass the name of a filedescriptor. This one is used to actually do the underlying read and write operations. However these functions get a ```fd``` argument passed. In some cases the descriptor passed to these makros actually _is_ ```fd```. This is not specific to the makros, the code was actually the same before.

Actually using ```fd``` are:

- qmail-pop3d.c
- qmail-popup.c
- qmail-smtpd.c

The argument is ignored and a hardcoded descriptor is used in:

- qmail-qmqpc.c
- qmail-remote.c

However I think that they are actually always identical, i.e. these functions could just use the ```fd``` argument and remove the makro argument entirely.

DerDakon:
Fixes #270.

There were situations where the generated functions would ignore the fd that was passed as their argument because if was always "-1". This happened when the substdio was statically initialized but a socket was dynamically allocated at runtime and was actually used for communication. Update the substdio when the socket is known, and then just use the passed fd.

### DIFF
--- a/timeoutwrite.h
+++ b/timeoutwrite.h
@@ -5,11 +5,11 @@
 
 extern ssize_t timeoutwrite(int t, int fd, const void *buf, size_t len);
 
-#define GEN_SAFE_TIMEOUTWRITE(funcname,tout,writefd,doexit) \
+#define GEN_SAFE_TIMEOUTWRITE(funcname,tout,doexit) \
 ssize_t funcname(int fd, const void *buf, size_t len) \
 { \
   ssize_t r; \
-  r = timeoutwrite(tout,writefd,buf,len); \
+  r = timeoutwrite(tout,fd,buf,len); \
   if (r == 0 || r == -1) doexit; \
   return r; \
 }
--- a/qmail-pop3d.c
+++ b/qmail-pop3d.c
@@ -24,8 +24,8 @@ void die() { _exit(0); }
 
 extern int rename(const char *, const char *);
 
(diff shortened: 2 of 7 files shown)


timeoutwrite.h
```diff
@@ -5,11 +5,11 @@
 
 extern ssize_t timeoutwrite(int t, int fd, const void *buf, size_t len);
 
-#define GEN_SAFE_TIMEOUTWRITE(funcname,tout,writefd,doexit) \
+#define GEN_SAFE_TIMEOUTWRITE(funcname,tout,doexit) \
 ssize_t funcname(int fd, const void *buf, size_t len) \
```
josuah, line 9: The `fd` argument was never used there. Maybe an omission or a change of mind mid-way...
This sounds like pursuing some work, makes sense to me.
Thank you!

josuah (approved):

timeoutwrite.h
DerDakon: Not entirely. Some callers passed "fd" as the "writefd" macro argument, so they actually used the fd passed to the function. And now just all do.

mbhangui (approved):

PR #279, opened 2024-05-28
Update NetBSD and Alpine, remove macOS 11.

schmonz:
A NetBSD 9.4 vmaction is available. Alpine now has a release containing riscv support. GitHub will be retiring macOS 11 in a month.

### DIFF
--- a/.github/workflows/check-available-os-versions.sh
+++ b/.github/workflows/check-available-os-versions.sh
@@ -1,0 +1,143 @@
+#!/bin/sh
+
+set -e
+
+SURPRISE_COUNT=0
+[ -z "${HTML2TEXT}" ] && HTML2TEXT=html2text
+
+fetch_webpage_as_markdown() {
+	curl -sS "$1" | ${HTML2TEXT}
+}
+
+fetch_alpine_upstream_latest_stable() {
+	fetch_webpage_as_markdown 'https://alpinelinux.org/downloads/' \
+		| grep 'Current Alpine Version' \
+		| awk '{print $4}' \
+		| sed -e 's|\*||g'
+}
+
+fetch_github_runner_list() {
+	fetch_webpage_as_markdown 'https://github.com/actions/runner-images' \
+		| perl -ne 'next unless /YAML Label/; print <>;' \
+		| perl -ne 'next if /^---/; print;' \
+		| perl -ne 'last if /Label scheme/; print;' \
+		| grep -v ' beta ' \
+		| grep -v ' deprecated ' \
+		| grep -v 'windows-' \
+		| cut -d'|' -f2 \
+		| sed -e 's| or |\n|g' -e 's|, |\n|g' -e 's|^ ||' -e 's|`||g' -e 's| $||' \
+		| grep -v -- '-latest' \
+		| grep -v -- '-large' \
+		| grep -v -- '-xlarge' \
+		| grep -v '^\s*$'
+}
+
+fetch_github_runner_os_versions() {
+	osname="$1"
+	osname_lower="$(echo ${osname} | tr '[:upper:]' '[:lower:]')"
+	fetch_github_runner_list \
+		| grep "^${osname_lower}-" \
+		| sed -e "s|^${osname_lower}-||" \
+		| tr '\n' ' ' \
+		| sed -e 's| $||'
+
+}
+
+fetch_vmactions_os_versions() {
+	osname="$1"
+	osname_lower="$(echo ${osname} | tr '[:upper:]' '[:lower:]')"
+	fetch_webpage_as_markdown "https://github.com/vmactions/${osname_lower}-vm" \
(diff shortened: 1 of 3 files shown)


josuah (approved):
Good thing that you are so close to VMActions!

.github/workflows/platform-builders.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: ["x86_64", "x86", "armhf", "armv7", "aarch64", "ppc64le", "riscv64", "s390x"]
-        osversion: ["edge"]
```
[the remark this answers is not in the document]
josuah: I though you did this change to point to a more stable version that does not risk to be updated at an unplanned time. "edge" might be adjusted at any time. A version is likely not going to evolve automatically.

DerDakon (approved):

schmonz:
@josuah @DerDakon could you take another look? I've added two more bits of automation to help our builds stay up-to-date:

1. Enabling Dependabot to report on any Actions we use that are outdated
2. Scripting OS version checks for the platforms we build on

With (2), if any platforms gain or lose available versions, the build will fail until something is done. If we don't want some new version, we can simply update the script. If we do want it, then we update `platform-builders.yml` as well.

(With (2), the build also fails if the script's scraping techniques fail. Might be annoying. Hopefully not often.)

With (1), I expect we get weekly notifications if any Action needs a version bump. If this turns out not to be enough automation for our needs, I'll add another shell script later :-)

.github/workflows/check-available-os-versions.sh
```diff
@@ -0,0 +1,108 @@
...
+	check_os_versions NetBSD 10.0 9.4 9.3 9.2 9.1 9.0
+	check_os_versions OpenBSD 7.5 7.4 7.3 7.2
+	check_os_versions Solaris 11.4-gcc 11.4
+	check_os_versions Ubuntu 22.04 20.04
+
+	return ${SURPRISE_COUNT}
+}
+
+main "$@"
+exit $?
```
josuah, line 108: I think this is implicit: the last command sets the exit status. Not a problem to keep it though

.github/workflows/check-available-os-versions.sh
```diff
@@ -0,0 +1,108 @@
...
+
+	check_os_versions Alpine 3.20.0
+	check_os_versions DragonFlyBSD 6.4.0
+	check_os_versions Fedora rawhide
+	check_os_versions FreeBSD 14.0 13.3 13.2 12.4
+	check_os_versions macOS 14 13 12
+	check_os_versions NetBSD 10.0 9.4 9.3 9.2 9.1 9.0
+	check_os_versions OpenBSD 7.5 7.4 7.3 7.2
+	check_os_versions Solaris 11.4-gcc 11.4
+	check_os_versions Ubuntu 22.04 20.04
```
josuah, line 102: Very nice and concise config-file like syntax, which I use a lot for scripts.
Would it be worth adding a comment on each .yaml file to tell "keep up to date with `check-available-os-versions.sh`"?

.github/workflows/check-available-os-versions.sh
```diff
@@ -0,0 +1,108 @@
...
+
+fetch_webpage_as_markdown() {
+	curl -sS "$1" | ${HTML2TEXT}
+}
+
+fetch_alpine_upstream_latest_stable() {
+	fetch_webpage_as_markdown 'https://alpinelinux.org/downloads/' \
+		| grep 'Current Alpine Version' \
+		| awk '{print $4}' \
+		| sed -e 's|*||g'
```
josuah, line 16: Is this really stripping `*`? I thought this would need to be escaped like `\*`.
No trouble with that though.

josuah (approved):
Adding more scripts inflates the complexity of the pipeline, but maybe it does not matter it this is an extra check that can be turned off without breaking anything like here. So more checks are one less thing to stab us in the back...

Scripting the automation scripts sounds much  admin-like. :)

.github/workflows/check-available-os-versions.sh
schmonz: Yeah, I've been following this shell script template (`set -e`, a `main()`, invoking it like this with an explicit `exit`) out of habit for years. It makes control flow clearer for people who don't write a lot of shell scripts, and when scripts get big it helps keep it clear even for the rest of us ;-)

.github/workflows/check-available-os-versions.sh
schmonz: Yes, I was surprised to see it work as is. Which is precisely why escaping is probably better. I'll test and commit if so.

josuah (approved):

josuah:
My bad, talked too fast... CI failing now. Maybe extra repositories required for `yq`.

Issue #86, opened 2019-08-26
LWQ (Life with qmail): update or fork


schmonz:
LWQ is the community's agreed best documentation for installing and running qmail. As we continue to diverge from qmail and netqmail, LWQ will become less accurate -- unless we are able to update it, or otherwise get it updated.

For instance, with notqmail 1.07, LWQ already doesn't describe installation from binary packages.

It's [possible to fork LWQ](http://www.lifewithqmail.org/lwq.html#lwq-license):

> _Life with qmail_ is covered by the OpenContent License, version 1.0. See http://www.opencontent.org/opl.shtml for the full license. Basically, you can copy, redistribute, or modify _Life with qmail_ provided that modified versions, if redistributed, are also covered by the OpenContent License.

But of course we'd rather not need to. And it's not at all inconceivable that Dave Sill would still be willing to update it and/or share write access with one or more of us.

schmonz:
Oops, never updated the status here.

As of September 2019, Henning Brauer (hosting provider for `lifewithqmail.org`) thought reviving qmail is a terrible idea and wanted neither to host updated docs nor to transfer the domain, while Dave Sill (author) was "happy to pass the LWQ torch to the notqmail team" and sent me the RCS (!!!) repo. We need a new name for it, at the very least, anyway. So we'll fork.

We've since sorted out one prerequisite: self-hosting [the website](https://notqmail.org). Here's what still needs doing.

1. Modernize [SDF](https://metacpan.org/dist/sdf), the source markup language:
    - Fix warnings and errors under recent Perl
    - Validate that we're generating identical output
    - Incorporate the fixes into [pkgsrc p5-sdf](https://github.com/NetBSD/pkgsrc/tree/trunk/textproc/p5-sdf)
    - Maybe issue a new SDF release to CPAN
2. Import the RCS repo into Git:
    - Probably by importing into CVS first
    - Attribute the Git commits to `lwq` at `sill.org`
    - Push it up to `/notqmail/lifewithqmail`, read-only, for posterity
3. Write an ikiwiki input format plugin for SDF
4. Import `lifewithqmail.git`'s history into a subdir of `notqmail.org`
5. Decide:
    - Whether it's trying to be more than one thing (e.g., a tutorial and a reference part)
    - Whether it still needs a name (or has simply become part of our documentation)
6. Start updating the docs:
    - Point non-notqmail users to `lifewithqmail.org`
    - Point archaeologists to `lifewithqmail.git`
[...]

PR #280, opened 2024-06-27
Retire DJB-era in-tree text docs.

schmonz:
We're still using CHANGES, README, and VERSION. All other DJB-era docs have been removed. Much of the information remains available (even updated!) from <https://cr.yp.to/qmail.html>, and the rest can always be extracted from <https://qmail.notqmail.org/netqmail-1.06.tar.gz>.


.github/workflows/check-available-os-versions.sh
josuah, line 1: That sounded unrelated, but nowhere harmful.

josuah (approved):
I agree that it is better have fewer documentation than out of date documentation.
What to point users at next? There is a link from the README to https://notqmail.org/ currently, which leads to https://notqmail.org/install/, would that work?

There is also https://qmail.notqmail.org/

And the manual pages in-tree.
Maybe https://github.com/notqmail/notqmail/blob/main/qmail.7 is also a good entry point.

.github/workflows/check-available-os-versions.sh
schmonz: The autobuild was failing because upstream Alpine reports "3.20.1". I tried bumping the Alpine version in `platform-builders.yml` to match, but the teeny version is apparently not part of the path for updating Alpine packages. So if teeny Alpine version bumps aren't going to be actionable for us, new teeny versions coming out shouldn't cause our autobuild to fail. That's this change.

schmonz:
> What to point users at next?

The README link to our website is probably sufficient for now. Maybe my imagination is poor, but I can't imagine someone looking at notqmail who doesn't know about "Life with qmail", DJB's website, etc.

Anyway, this is a very good reminder to make more progress on #86.

.github/workflows/check-available-os-versions.sh
josuah: I was confident there were good reasons for it, but now I know why!
Thank you!

PR #282, opened 2024-12-21
commands: decouple command processing and IO

petershh:
This would allow adding timed reads in pop3d, popup and smtpd

Also add custom data pointer to command handlers to faciliate getting rid of global variables.

Along the way, get rid of K&R function definitions in the related files.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -5,38 +5,26 @@
 #include "str.h"
 #include "case.h"
 
-static stralloc cmd = {0};
-
-int commands(ss,c)
-substdio *ss;
-struct commands *c;
+void execute_command(char *line, size_t len, const struct command *cmds)
 {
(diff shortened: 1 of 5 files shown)


DerDakon (changes requested):
You are touching many functions, even some maybe unrelated ones. That is no problem, but please split out all these "get rid of K&R" stuff in it's own commit, likely the first one in your series. Then add the API adaptions that you need, and please use ```NULL``` instead of ```0``` for pointers there (```okay(0, 0);``` popped out, but there may be others). Also please add ```const``` to arguments where appropiate, especially for the ```char*``` ones of put/print functions. And then do the actual changes as it's own commit, this hopefully makes all of this easier to review. Thanks!

PR #291, opened 2026-04-20
quarterly autobuild update

schmonz:
Bump FreeBSD and macOS; switch OmniOS to the version that includes build-essentials already.

We've been avoiding some updates to platforms with gcc14 or newer. Would be good to try passing `--std=foo` to see if that lets us get on the latest. (And if not, that would lend some urgency to driving down the rest of the warnings.)

### DIFF
--- a/.github/workflows/check-available-os-versions.sh
+++ b/.github/workflows/check-available-os-versions.sh
@@ -124,7 +124,7 @@ get_known_platform_versions() {
 	Alpine) echo 3.23 ;;
 	DragonFlyBSD) echo 6.4.2 6.4.1 ;;
 	Fedora) echo rawhide ;;
-	FreeBSD) echo 15.0 14.3 14.2 14.1 14.0 13.5 13.4 13.3 13.2 ;;
+	FreeBSD) echo 15.0 14.4 14.3 14.2 14.1 14.0 13.5 13.4 13.3 13.2 ;;
 	macOS) echo 26 15-intel 15 14 ;;
 	NetBSD) echo 10.1 10.0 9.4 9.3 9.2 9.1 9.0 ;;
 	OmniOS) echo r151056 r151054 r151052 r151050 r151048 r151046 ;;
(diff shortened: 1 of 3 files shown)


.github/workflows/check-available-os-versions.sh
```diff
@@ -124,10 +124,10 @@ get_known_platform_versions() {
 	Alpine) echo 3.23 ;;
 	DragonFlyBSD) echo 6.4.2 6.4.1 ;;
 	Fedora) echo rawhide ;;
-	FreeBSD) echo 15.0 14.3 14.2 14.1 14.0 13.5 13.4 13.3 13.2 ;;
-	macOS) echo 26 15-intel 15 14 ;;
+	FreeBSD) echo 15.0 14.4 14.3 14.2 14.1 14.0 13.5 13.4 13.3 13.2 ;;
+	macOS) echo 26 26-intel 15-intel 15 14 ;;
 	NetBSD) echo 10.1 10.0 9.4 9.3 9.2 9.1 9.0 ;;
-	OmniOS) echo r151056 r151054 r151052 r151050 r151048 r151046 ;;
+	OmniOS) echo r151056-build r151056 r151054 r151052 r151050 r151048 r151046 ;;
```
josuah, line 130: It seems like a version of OmniOS with build-essential preinstalled.
https://github.com/vmactions/omnios-vm/issues/8

Good for reducing CI workload.

.github/workflows/platform-builders.yml
josuah, line 214: Did you wish t put `-build` here too? I think the comment you did was "note but do not upgrade" as in "do not switch to build-essential" just yet?

josuah (approved):
Thank you for the maintenance!

DerDakon (approved):

.github/workflows/check-available-os-versions.sh
schmonz: Oooh, thanks for following up on this. Exactly what we need. Changed the branch to use it.

.github/workflows/platform-builders.yml
schmonz: I hadn't, but now that I know what it contains (thanks!), I have now.

PR #292, opened 2026-04-23
Autobuild: give sufficient perms for CodeQL analyse.

schmonz:
Previous autobuild update defined missing permissions, but too tightly. This should fix the CodeQL analysis job, and in turn that should resolve the 11 "Security and quality" items.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,5 +1,7 @@
 name: "CodeQL"
-permissions: read-all
+permissions:
+  contents: read
+  security-events: write
 
 on:
   push:


josuah (approved):
It seems like this gives CodeQL the ability to read security events, however this is an open-source + transparent project AFAIU, so no impact.

https://github.com/orgs/community/discussions/29710#discussion-4309566

Thank you!

schmonz:
> It seems like this gives CodeQL the ability to read security events

Yes, and thank you again for finding and linking the most needed context. My understanding is that this is no worse than before the new permissions system landed, and is merely more explicit -- a good chance to review the existing risk, and maybe pile on that linked thread to agitate for improvement ;-)

DerDakon (approved):

PR #294, opened 2026-09-10
    refactor(qbiff)!: remove qbiff utility and utmp legacy code

JustCommitRandomness:
### Summary
The project has been fearless in removing auxiliary diagnostic programs; this PR proposes that `qbiff` join them, completely eliminating the legacy `utmp` to `utmpx` shim.

### Rationale
* **biff n** is the modern system default.
* **Platform Availability:** Modern Linux distributions tend to come without `biff` installed or packaged out of the box. 
* **Changing Architectures:** It is rare for email to be delivered to mailboxes corresponding to local login accounts, and rare for a mail server to act as a daily driver CLI environment.
* **Modern Alternatives:** Almost everyone uses permanently active mail clients, IMAP NOTIFY applets, or notifications via `notify-send` (libnotify), which provide more flexible alternatives.

In my opinion, the subset of users who actually wish to be interrupted via terminal injection just because an email has arrived has shifted from rare to pure fiction; I'd be surprised if `|qbiff` appears in any`.qmail` files written in the last 10 years

### Technical Debt Clean-up
This thought was inspired by the `notqmail` implementation having been restructured for `utmpx.h` with a `utmpx` to `utmp` shim hidden in `qtmp.h1`. 

Looking at the opaque `#define` horror preserving the original DJB `utmp.h` code, removing the implementation entirely is a far cleaner path than addressing the incompatible pointer warnings. This eliminates the dynamic `tryutmpx.c` runtime compilation check and strips out an implicit dependency on fragile compiler flag padding assumptions matching the host operating system's struct layout.

### Impact & Breaking Changes
* This completely removes the `qbiff` binary, documentation (`qbiff.1`, `qbiff.0`), and its configuration artifacts (`qtmp.h1`, `qtmp.h2`, `tryutmpx.c`).
* This will not affect email sending or delivery, but may stop someone receiving a notification that they received new email on their IBM3151. [^1]
* Cleans up installation pathways in `hier.c` and global documentation cross-references in `qmail.7`.

[^1]: I administer my mailserver via an IBM3151, I do not use qbiff.


schmonz:
Happy to see a new contributor. Thanks for making the effort!

I'm curious what's motivating this change. I mean clearly qbiff is of dubious utility in typical 2026 mail server configs, but besides generally wanting to remove code that's not worth the carrying cost, what's prompting you to want to remove this particular code at this particular time? Have you found it to be particularly costly?

Besides your answer to that, I'd also like to know by what means you prepared and tested these changes. We don't as yet have a policy for LLM-assisted contributions, but whatever it will turn out to be starts with knowing whether this is one.

JustCommitRandomness:
I found myself needing a new mailserver, and after a couple of weeks of looking to see what the current crop were like, still deciding I wanted to stay with qmail, But as  this server will be 64 bit, not 32 bit, I was looking for a refreshed codebase. s\qmail has too much crammed into it. Similarly I looked at nosh but ended up staying with s6.

I digress.

The incompatible pointer warning in qtmp.h1 made me look at it. Rather than just throwing a (void *) in I started trying to work out what the code was doing, and those macros make it really hard to read the code... unless you're on a utmpx system.

If I'd started off looking at the original code, or at qtmp.h2, I'd have got it much quicker, but I spent way too much time trying to work it out, and then I wondered what the heck qmail needed to be doing things with utmp for anyway... and then I realised I wasted some hours on a flipping biff replacement, thew my hands up in the air and rage deleted it, Then I though I could spare some other poor soul from the BSD wastelands having to deal with the same issues.

JustCommitRandomness:
uhh. testing. I grepped qtmp.h and only qbiff was involved. I grepped qbiff and made the changes. I made sure it still built and installed. Then I undid the changes to Thoughts.md, because those were historical not current

unfortunately, I did all this from my deployment branch, which has patches merged in, so I had to faff around with git to apply it to the base install and force push, but the I realised that even then I was working from the last release, not the current tree, so I did it again, and generally got myself confused with more branches than I can shake a twig at.

DerDakon:
You also need to remove the utmp-specific build jobs. Otherwise I agree that this code doesn't serve any real purpose anymore.

PR #295, opened 2026-09-12
remove `qbiff` and its entrails

theAlexes:
`qbiff` hails from the `mbox` era and is unlikely to be useful on a modern system with Maildir delivery and no shell logins. Furthermore, the original author of `qmail` considers its requisite `comsat` daemon to be abominable. `qbiff` also necessitated some code to interact with the `utmp` file, which can all be removed now.

This is a breaking change. `.qmail` files will need to be updated to remove any existent `|qbiff` lines.


DerDakon:
Same as #294 .

